PR #1, opened 2021-08-20
Sandbox Runtime

ChaoticTempest:
Rough implementation of `SandboxRuntime`: not everything is in place, but works for the most part in testing things locally. Also, to make it more idiomatic to rust's runtime libraries, there's no concept of `Runner` implemented, since `Runtime` is basically the runner.

- After this PR, will rename the repo to `runner-rs`.
- All the helper functions got moved to `runner/rpc/api.rs`, but will likely be changed over to use NAR after this PR gets merged.
- `SandboxRuntime` has been added, but `mainnet` and `testnet` runtimes are not implemented yet (should be easy in a future ticket to implement and hopefully not "famous last words").

Sample usage:
```rust
use runner::*;

#[runner::test(sandbox)]
async fn test_deploy() {
    let (contract_id, signer) = dev_deploy(Path::new("path/to/file.wasm"))
        .await
        .expect("could not dev-deploy contract");

    let result = view(
        contract_id,
        "function_name".to_string(),
        r#""some_arg": "some_value""#.into(),
    ).await.expect("could not call into view function");

    assert_eq!(result, OUR_EXPECTED_RESULT);
}

```

or using it within a binary directly like so:
```rust
use runner::*;

#[runner::main(sandbox)]
async fn main() {
}
```
The hope is we can have `#[runner::test(testnet)]` and `#[runner::test(mainnet)]` options too, which do work, but just the runtime hasn't been implemented for those.

Let me know if there's anything anybody doesn't understand. The design for the runtime was mainly taken from `tokio` itself.


runner-macros/src/entry.rs
```diff
@@ -0,0 +1,148 @@
...
+                    .to_string()
+                    .to_lowercase();
+
+                let msg = format!("Not expected to received a value for: {}", ident.as_str());
+                return Err(syn::Error::new_spanned(namevalue, msg));
+            }
+            syn::NestedMeta::Meta(syn::Meta::Path(path)) => {
+                let name = path
+                    .get_ident()
+                    .ok_or_else(|| syn::Error::new_spanned(&path, "Must have specified ident"))?
```
austinabell, line 40: Why not default to `sandbox` so that a user can just write `#[runner::test]`?

runner-macros/src/entry.rs
```diff
@@ -0,0 +1,148 @@
...
+        match arg {
+            syn::NestedMeta::Meta(syn::Meta::NameValue(namevalue)) => {
+                let ident = namevalue
+                    .path
+                    .get_ident()
+                    .ok_or_else(|| {
+                        syn::Error::new_spanned(&namevalue, "Must have specified ident")
+                    })?
+                    .to_string()
+                    .to_lowercase();
```
austinabell, line 32: why convert to lowercase if this branch will always error?

runner-macros/src/entry.rs
```diff
@@ -0,0 +1,148 @@
...
+        // `Span` on stable Rust has a limitation that only points to the first
+        // token, not the whole tokens. We can work around this limitation by
+        // using the first/last span of the tokens like
+        // `syn::Error::new_spanned` does.
+        let start = last_stmt.next().map_or_else(Span::call_site, |t| t.span());
+        let end = last_stmt.last().map_or(start, |t| t.span());
+        (start, end)
+    };
+
+    let rt = match flavor.unwrap() {
```
austinabell, line 86: Perhaps this should also allow configuring through env variable? Can just open an issue to tackle later, but would be much more preferable to just set an env variable to override where the tests are run instead of forcing code changes to test against diff networks.

runner-js has `NEAR_RUNNER_NETWORK`

runner/src/runtime/context.rs
```diff
@@ -0,0 +1,32 @@
...
+    static RT_CONTEXT: RefCell<Option<RuntimeFlavor>> = RefCell::new(None);
+}
+
+pub(crate) fn current() -> Option<RuntimeFlavor> {
+    RT_CONTEXT.with(|ctx| ctx.borrow().clone())
+}
+
+pub(crate) fn enter(flavor: RuntimeFlavor) -> EnterGuard {
+    RT_CONTEXT.with(|ctx| {
+        let old = ctx.borrow_mut().replace(flavor);
```
austinabell, line 14: I'm a little confused by this, because if used asynchronously wouldn't this cause race conditions where another thread/task would access the modified environment rather than the intended one?

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,149 @@
...
+        Self {
+            port,
+            process: None,
+        }
+    }
+
+    pub fn new_default() -> Self {
+        let port = pick_unused_port().expect("no ports free");
+        Self::new(port)
+    }
```
austinabell, line 36: Just implement `Default` instead of this

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,149 @@
...
+#[cfg(not(target_os = "windows"))]
+fn init_sandbox(home_dir: &Path) -> io::Result<Child> {
+    Command::new(
+        "/usr/local/lib/node_modules/near-sandbox/node_modules/binary-install/bin/near-sandbox",
+    )
+    .arg("--home")
+    .arg(home_dir)
+    .arg("init")
+    .spawn()
+}
```
austinabell, line 120: Can you merge the function signatures so only the inner functionality is separated by `cfg`? This just opens up to potential issues with function signatures not matching if something changed and can go easily unnoticed.

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,149 @@
...
+use std::io;
+use std::path::{Path, PathBuf};
+use std::process::{Child, Command};
+use std::{thread, time::Duration};
+
+use super::context;
+use super::RuntimeFlavor;
+
+fn local_rpc_addr(port: u16) -> String {
+    format!("0.0.0.0:{}", port)
```
austinabell, line 12: why invalid address?

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,149 @@
...
+            _guard: context::enter(RuntimeFlavor::Sandbox(port)),
+        }
+    }
+
+    pub fn port(&self) -> u16 {
+        self.server.port
+    }
+
+    pub fn rpc_addr(&self) -> String {
+        format!("http://localhost:{}", self.port())
```
austinabell, line 143: This isn't used, so I'm curious why add it now?

runner/src/rpc/api.rs
```diff
@@ -230,68 +228,8 @@ pub async fn dev_create() -> Result<(AccountId, InMemorySigner), String> {
 
 pub async fn dev_deploy(contract_file: &Path) -> Result<(AccountId, InMemorySigner), String> {
```
austinabell, line 229: Don't necessarily need to worry about generic usage for this so can use `impl AsRef<Path>` to allow this to be called like `dev_deploy("<some-path>")` and it'll remove boilerplate from tests

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,149 @@
...
+impl SandboxRuntime {
+    pub fn new_default() -> Self {
+        let server = SandboxServer::new_default();
+        let port = server.port;
+
+        Self {
+            server,
+            _guard: context::enter(RuntimeFlavor::Sandbox(port)),
+        }
+    }
```
austinabell, line 136: impl `Default` instead

runner/tests/deploy.rs
```diff
@@ -0,0 +1,44 @@
...
+        None,
+    )
+    .await
+    .unwrap();
+
+    tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
+    let call_result = view(
+        contract_id.clone(),
+        "nft_metadata".to_string(),
+        b"".to_vec().into(),
```
austinabell, line 37: Optional:
```suggestion
        Vec::new().into(),
```

runner/tests/deploy.rs
```diff
@@ -0,0 +1,44 @@
...
+        contract_id.clone(),
+        "new_default_meta".to_string(),
+        format!("{{\"owner_id\": \"{}\"}}", contract_id).into(),
+        None,
+    )
+    .await
+    .unwrap();
+
+    tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
+    let call_result = view(
```
austinabell, line 34: This view function returns a `serde_json::Value`, but it would be much more useful if it deserialized into a generic type of the caller's choice. This way you don't need to re-serialize the result to compare against a static string and can compare actual rust types.

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,149 @@
...
+fn init_sandbox(home_dir: &Path) -> io::Result<Child> {
+    Command::new("near-sandbox")
+        .args(&["--home", home_dir, "init"])
+        .spawn()
+}
+
+#[cfg(not(target_os = "windows"))]
+fn start_sandbox(home_dir: &Path, port: u16) -> io::Result<Child> {
+    Command::new(
+        "/usr/local/lib/node_modules/near-sandbox/node_modules/binary-install/bin/near-sandbox",
```
austinabell, line 101: This is broken because it's your fs specific. Why not just use the near-sandbox binary and expect that it's in the user's path for now?

austinabell (changes requested):
tests don't work locally because we have diff sandbox binary locations. I'll dig more into the internals when I can test things by running

runner-macros/src/entry.rs
ChaoticTempest: Good idea. I was thinking something like that too, but didn't immediately add that. Let me see if I can't just put it in

runner-macros/src/entry.rs
ChaoticTempest: oops, thanks for noticing this. This macro was mainly copypasta'd over from tokio

runner-macros/src/entry.rs
ChaoticTempest: 👍 Will make that an issue to tackle later

runner/src/runtime/context.rs
ChaoticTempest: This was also another thing I took from tokio. This is behind a `thread_local` so using it asynchronously shouldn't be an issue. You can also take a look at how tokio uses it [context::enter](https://github.com/tokio-rs/tokio/search?q=context%3A%3Aenter)

There is another case where the runtime is multi-threaded, but the runtime will do the thread spawning directly instead of the developer. During thread spawning we also call `enter` for that specific thread to have this runtime context as well. But the current rt implementation is only single threaded.

ChaoticTempest:
> tests don't work locally because we have diff sandbox binary locations. I'll dig more into the internals when I can test things by running

@austinabell Ahh, yea sorry about that. Just hardcoded it for now (base on my binary location) so there's proper cleanup after the test finishes. We just need to have a cargo crate of the sandbox binary or a utility that can easily fetch the sandbox location if we install it thru npm.

runner/src/runtime/local.rs
ChaoticTempest: hmm, I just used the same address in runner-js and this gets used by `near-sandbox` itself. Is it throwing invalid address for you?

runner/src/runtime/local.rs
ChaoticTempest: It probably was used at the start, but did some cleanup and didn't get used anymore. Nice catch, I'll clean it up too. Thanks

runner/src/runtime/local.rs
ChaoticTempest: Same thing I mentioned in the global discussion. But I should mention that `near-sandbox` is just a script that launches the internal near-sandbox binary. Killing the process only kills the scripts so I needed the direct path

ChaoticTempest: also, you can get your own specific binary location with
```
near-sandbox run
ps -ef | grep near-sandbox
```
if you need it

examples/src/nft.rs
```diff
@@ -0,0 +1,67 @@
...
+    println!("new_default_meta outcome: {:#?}", outcome);
+
+    tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
+    let deposit = 10000000000000000000000;
+    let outcome = call(
+        &signer,
+        contract_id.clone(),
+        contract_id.clone(),
+        "nft_mint".to_string(),
+        format!(
```
willemneal, line 34: I thinking using serde's `json!` macro makes this easier to read and has the benefit of pulling in value in the current scope.

runner/src/runtime/local.rs
austinabell: ah, didn't realize sandbox accepted this to let it auto-assign

examples/src/nft.rs
ChaoticTempest: done

runner/src/rpc/api.rs
```diff
@@ -97,7 +95,7 @@ pub async fn view(
     method_name: String,
     args: FunctionArgs,
 ) -> Result<serde_json::Value, String> {
```
austinabell, line 97: Can we swap this to be generic on serde deserialize to not have to go through serde_json::Value if a caller doesn't want? (prob in another PR)

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,141 @@
...
+                "--home",
+                home_dir.to_str().unwrap(),
+                "run",
+                "--rpc-addr",
+                &local_rpc_addr(port),
+            ])
+            .spawn()
+    } else {
+        Command::new(
+            "/usr/local/lib/node_modules/near-sandbox/node_modules/binary-install/bin/near-sandbox",
```
austinabell, line 93: Can you open an issue to fix these? Seems really bad to have this assumption. I remember seeing changes for this to use an installed version, was that not here?

runner/src/runtime/local.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl Drop for SandboxServer {
+    fn drop(&mut self) {
+        if self.process.is_none() {
+            return;
+        }
+
+        let child = self.process.as_mut().unwrap();
+
+        println!(
```
austinabell, line 67: Might want these to be `eprintln`

austinabell (approved):
Seems like a decent starting point, as long as the install and binary usage is updated, sleeps are updated to not be so rigid, and certain interfaces are cleaned up in another PR.

Definitely want to make this accessible for anyone to use without assuming a specific binary location of something installed externally to test this

runner/src/rpc/api.rs
ChaoticTempest: yea definitely. A lot of the RPC calls need to be re-evaluated too. I'll make a ticket for another PR later once this merges

runner/src/runtime/local.rs
ChaoticTempest: I moved these over to the `near-sandbox-utils` package instead. So will update these once https://github.com/near/sandbox/pull/12 merges

runner-macros/src/entry.rs
```diff
@@ -0,0 +1,152 @@
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::{quote, quote_spanned, ToTokens};
+
+enum Flavor {
```
willemneal, line 5: Why not network?

willemneal (approved):
Approved, just have one comment about the type name `flavor` over `network` or even `endpoint`.

runner-macros/src/entry.rs
ChaoticTempest: The `RuntimeFlavor` currently doesn't have much info besides port for sandbox, but I think eventually would have a bit more info related to the context of the runtime like if its multi-threaded or not. So not necessarily just containing network info which is why I think flavor works better since its more general

willemneal: But each one is an endpoint or network.  We can always refactor later.

ChaoticTempest: yea definitely. It's an internal struct anyways so we can change it to whatever

examples/src/nft.rs
```diff
@@ -0,0 +1,66 @@
...
+
+const NFT_WASM_FILEPATH: &str = "./examples/res/non_fungible_token.wasm";
+
+#[runner::main(sandbox)]
+async fn main() {
+    let (contract_id, signer) = dev_deploy(Path::new(NFT_WASM_FILEPATH)).await.unwrap();
+
+    // Wait a few seconds for create account to finalize:
+    // TODO: exponentialBackoff to not need these explicit sleeps
+    tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
willemneal, line 14: Could you add this to `dev_deploy`, etc for now?

PR #6, opened 2021-10-01
feat: Add sandbox view-state and patch-state

ChaoticTempest:
This adds patch-state and view-state function calls into sandbox. This works towards getting up a spooning example (mentioned in https://github.com/near/near-sdk-rs/issues/535) once testnet runtime is implemented.


runner/src/rpc/tool.rs
```diff
@@ -96,3 +107,18 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
     path.push(format!("{}.json", account_id));
     Ok(path)
 }
+
+/// Convert `StateItem`s over to a Map<data_key, value_bytes> representation.
+/// Assumes key and value are base64 encoded, so this also decode them.
+pub(crate) fn into_state_map(state_items: Vec<StateItem>) -> HashMap<String, Vec<u8>> {
+    let decode = |s: StateItem| {
+        (
+            str::from_utf8(&base64::decode(s.key.clone()).unwrap())
```
austinabell, line 116:
```suggestion
            str::from_utf8(&base64::decode(&s.key).unwrap())
```

runner/src/rpc/tool.rs
```diff
@@ -96,3 +107,18 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
     path.push(format!("{}.json", account_id));
     Ok(path)
 }
+
+/// Convert `StateItem`s over to a Map<data_key, value_bytes> representation.
+/// Assumes key and value are base64 encoded, so this also decode them.
+pub(crate) fn into_state_map(state_items: Vec<StateItem>) -> HashMap<String, Vec<u8>> {
+    let decode = |s: StateItem| {
```
austinabell, line 114: why not have this closure return a result so that it doesn't panic on invalid response?

runner/src/rpc/tool.rs
```diff
@@ -96,3 +107,18 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
...
+        (
+            str::from_utf8(&base64::decode(s.key.clone()).unwrap())
+                .unwrap_or_else(|_| &s.key)
```
austinabell, line 117: Wait, you want to use the key if the base64 decoded bytes aren't utf-8? This is all really confusing, if this is the case, can you document the specifics around why these specific semantics are required?

runner/src/rpc/tool.rs
```diff
@@ -96,3 +107,18 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
     path.push(format!("{}.json", account_id));
     Ok(path)
 }
+
+/// Convert `StateItem`s over to a Map<data_key, value_bytes> representation.
+/// Assumes key and value are base64 encoded, so this also decode them.
+pub(crate) fn into_state_map(state_items: Vec<StateItem>) -> HashMap<String, Vec<u8>> {
```
austinabell, line 113: Extension on the previous point, this should really return a result, since this is very possible to panic

runner/src/rpc/tool.rs
```diff
@@ -96,3 +107,18 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
...
+    let decode = |s: StateItem| {
+        (
+            str::from_utf8(&base64::decode(s.key.clone()).unwrap())
+                .unwrap_or_else(|_| &s.key)
+                .to_owned(),
+            base64::decode(s.value).unwrap(),
+        )
+    };
+
+    state_items.into_iter().map(decode).collect()
```
austinabell, line 123: You shouldn't need owned values for anything in the decode function, perhaps this would be more generally usable if the function just took a slice of elements and this just iterated through them?

runner/src/rpc/tool.rs
```diff
@@ -96,3 +107,18 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
     path.push(format!("{}.json", account_id));
     Ok(path)
 }
+
+/// Convert `StateItem`s over to a Map<data_key, value_bytes> representation.
+/// Assumes key and value are base64 encoded, so this also decode them.
```
austinabell, line 112:
```suggestion
/// Assumes key and value are base64 encoded, so this also decodes them.
```

runner/src/rpc/api.rs
```diff
@@ -119,6 +129,54 @@ pub async fn view(
...
+                account_id: contract_id,
+                prefix: prefix.unwrap_or_else(|| vec![].into()),
+            },
+        })
+        .await
+        .map_err(|err| format!("Failed to query state: {:?}", err))?;
+
+    match query_resp.kind {
+        QueryResponseKind::ViewState(state) => Ok(tool::into_state_map(state.values)),
+        _ => Err("Could not retrieve access key".to_owned()),
```
austinabell, line 149: Isn't this an inaccurate error, because this will just be hit if you get an unexpected response?

runner/src/rpc/api.rs
```diff
@@ -119,6 +129,54 @@ pub async fn view(
...
+    }
+}
+
+pub async fn patch_state<T>(
+    account_id: AccountId,
+    key: String,
+    value: T,
```
austinabell, line 156:
```suggestion
    value: &T,
```

This just needs to be serialized, should just pass references

runner/src/rpc/api.rs
```diff
@@ -119,6 +129,54 @@ pub async fn view(
...
+
+    match query_resp.kind {
+        QueryResponseKind::ViewState(state) => Ok(tool::into_state_map(state.values)),
+        _ => Err("Could not retrieve access key".to_owned()),
+    }
+}
+
+pub async fn patch_state<T>(
+    account_id: AccountId,
+    key: String,
```
austinabell, line 155: This feels like a weird signature for patch state since if someone wants to patch multiple keys, they need to send a request for every single modification when the API supports patching in bulk. I really think this should take an `IntoIterator` where each item is a K, V. If you still want the convenience of a simpler API for one element, maybe another function could be considered which just uses this with [`iter::once`](https://doc.rust-lang.org/std/iter/fn.once.html) under the hood?

austinabell (changes requested):
Some initial thoughts and questions

runner/src/rpc/api.rs
willemneal: In the js version we use a `RecordBuilder` to aggregate changes for a specific account.  But still have this API for a single update.

examples/src/status_message.rs
```diff
@@ -0,0 +1,44 @@
...
+        None,
+    )
+    .await
+    .unwrap();
+
+    let result = runner::view(
+        contract_id.clone(),
+        "get_status".into(),
+        json!({
+            "account_id": contract_id.clone().to_string(),
```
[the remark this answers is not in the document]
ChaoticTempest: Oh good catch, I have get-it-to-work-itis for this PR, so thanks for catching all of these rough edges

runner/src/rpc/api.rs
ChaoticTempest: @austinabell This ones actually tricky: taking an `IntoIterator` doesn't work for this case since we can't make a trait object out of `BorshSerialize` since it reference `self` in its methods. We can do `IntoIterator<Item=(Key, impl BorshSerialize)>` but now all patches have to be the same type in each iterable. I guess we'd end up needing something like what @willemneal suggested with the `RecordBuilder` potentially, but that requires exposing the developer to a lower level API, dealing directly with `StateRecord`s. I'm not sure what's the best way of handling this one

runner/src/rpc/tool.rs
ChaoticTempest: Originally when I was playing around with it, the keys looked weird and I couldn't seem to decode any of them, which is why there was this hackish mess. Thanks for pointing it out. I changed it to just return an `anyhow` result instead

runner/src/rpc/tool.rs
ChaoticTempest: That's a good point, unnecessary ownership is equally bad here. Fixed

runner/src/rpc/api.rs
ChaoticTempest: yup, you're right on this one. This was just copy-pasta from something else

runner/src/rpc/api.rs
austinabell: Good point, probably use cases will require different value types. Perhaps the API is just `Item=(Key, Vec<u8>)` for the iterator and then functionality can be built on top of this, and you can keep the same signature for this patch single state function? We probably want to think of the best lower-level API, then think about what exactly to expose in this case.

ChaoticTempest: let's just keep the single state patch for now then. I'll make a separate ticket to address dealing the rest of this lower level API. Sounds good to you?

runner/src/rpc/tool.rs
```diff
@@ -109,16 +109,16 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
...
-    let decode = |s: StateItem| {
-        (
-            str::from_utf8(&base64::decode(s.key.clone()).unwrap())
-                .unwrap_or_else(|_| &s.key)
-                .to_owned(),
-            base64::decode(s.value).unwrap(),
-        )
+/// Assumes key and value are base64 encoded, so this also decodes them.
+pub(crate) fn into_state_map(
+    state_items: &Vec<StateItem>,
```
austinabell, line 114: should be `&[StateItem]` since you don't use any vec specific things

runner/src/rpc/tool.rs
```diff
@@ -109,16 +109,16 @@ pub(crate) fn credentials_filepath(account_id: AccountId) -> Result<PathBuf, Str
...
+) -> anyhow::Result<HashMap<String, Vec<u8>>> {
+    let decode = |s: &StateItem| {
+        Ok((
+            str::from_utf8(&base64::decode(&s.key)?)?.to_owned(),
```
austinabell, line 118: Can use https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8 here

austinabell (changes requested):

PR #10, opened 2021-10-12
Testnet runtime

ChaoticTempest:
This adds the testnet runtime which is parallel to the sandbox runtime. All tests will work in `runner/tests` if we replace `runner::test(sandbox)` with `runner::test(testnet)` instead. Similarly, `examples/` will work too.

https://github.com/near/runner-rs/pull/6 should be merged into main first, then this one can be merged


volovyk-s:
@ChaoticTempest I do not have any objections except those mentioned in the meeting. Let's fix them and move forward.

PR #15, opened 2021-10-15
chore: Rename runner to workspaces

ChaoticTempest:
Since repo got renamed, the code needs to be updated as well


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 members = [
-    "runner",
-    "runner-macros",
+    "workspaces",
+    "workspaces-macros",
```
austinabell, line 4: Might want to squat these on crates.io, they don't currently exist but good to be sure

ChaoticTempest: good call. Should we go with purely `workspaces` or `near-workspaces`?

austinabell: idk I like the simplicity of `workspaces` but probably best to get others opinions if it would be better to have explicit near naming

cc @DiscRiskandBisque @jberrytech

Issue #13, opened 2021-10-13
Setup GitHub actions for this repo


ChaoticTempest:
Need to setup some basic CI for this repo, since a couple of things can get thru the cracks, such as formatting and unused imports. Can just be GitHub actions for now since that seems to be the simplest

PR #17, opened 2021-10-19
feat: Github actions workflow

ChaoticTempest:
Addresses https://github.com/near/workspaces-rs/issues/13.

This adds a pipeline where clippy will be ran first, then if succeeds, moves onto running `cargo test`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,21 +4,27 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
   check:
+    name: Clippy and fmt
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - run: rustup component add clippy
-      - uses: actions-rs/clippy-check@v1
+      - uses: actions/checkout@v2
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features
+          profile: minimal
+          toolchain: stable
+          components: rustfmt, clippy
+      - uses: Swatinem/rust-cache@v1
+      - name: Check Formatting
+        run: cargo fmt -- --check
+      - name: Check Clippy
+        run: cargo clippy --features unstable --tests -- -Dclippy::all
 
   test:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Check Formatting
+        run: cargo fmt -- --check
+      - name: Check Clippy
+        run: cargo clippy -- -Dclippy::all
+
+  test:
+    needs: check
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
```
volovyk-s, line 33: Can we add Windows here?
(we can skip it if it's tricky)

volovyk-s (approved):
LGTM!

.github/workflows/test.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt, clippy
+      - uses: Swatinem/rust-cache@v1
+      - name: Check Formatting
+        run: cargo fmt -- --check
+      - name: Check Clippy
+        run: cargo clippy -- -Dclippy::all
```
austinabell, line 27:
```suggestion
        run: cargo clippy --tests -- -Dclippy::all
```

Pretty optional but can help make tests more readable sometimes to make sure there aren't weird patterns followed

.github/workflows/test.yml
austinabell: it can easily be added to CI, but not so sure it is windows compatible as of now

.github/workflows/test.yml
```diff
@@ -0,0 +1,41 @@
...
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
+
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --verbose
```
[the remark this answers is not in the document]
austinabell: yes, it does

.github/workflows/test.yml
ChaoticTempest: I don't think we have windows binaries for sandbox yet unfortunately. @chefsale I remember Willem was talking to you about those. Are there any new updates on this?

Issue #7, opened 2021-10-04
Decide on and switch error handling to something more generally usable


austinabell:
Currently just using string errors for everything, but would probably be best for long-term maintainability if we pick an error type/pattern to use before the API surface gets large and starts to be used.

Options:
- Concrete types that implement `std::error::Error`
  - Probably too tedious to have many error types for this kind of API
  - Can use https://github.com/dtolnay/thiserror to derive for convenience on impls
- `Box<dyn std::error::Error>` error types
  - Reduces code internally to avoid having to convert errors from other crates, but can be annoying to use in cases
- Anyhow Error https://github.com/dtolnay/anyhow
  - probably just a better version of the above, doesn't seem like it adds any overhead to a user other than potentially being a bit confusing to use for new Rust devs, but probably not much more than `String` errors
- Keep `String` error on public APIs, use one of the above internally for convenience

I personally think using something like anyhow makes sense for this kind of API that is just used for testing, but let me know if you guys have thoughts. Concrete error types could also be an option but this might become tedious without any real benefit

ChaoticTempest:
`anyhow` seems to be a good match for this so far. Very thin and can easily swap between errors that implement `std::error::Error`. Anyhow also offers downcasting to a more specific error, so a lot better than purely just handling a `String` and trying to grab its contents. Let's just swap over `anyhow` then. It's also already being used somewhat internally

austinabell:
> `anyhow` seems to be a good match for this so far. Very thin and can easily swap between errors that implement `std::error::Error`. Anyhow also offers downcasting to a more specific error, so a lot better than purely just handling a `String` and trying to grab its contents. Let's just swap over `anyhow` then. It's also already being used somewhat internally

using boxed std errors [can also be downcasted](https://doc.rust-lang.org/std/error/trait.Error.html#method.downcast) but yeah in my anecdotal experience using these boxed errors is not ideal and anyhow is just a cleaner alternative in almost every way.

austinabell:
@ChaoticTempest we can close this now, or do you think this is still unfinished?

ChaoticTempest:
Definitely can close this now. Thanks for the reminder

PR #20, opened 2021-11-08
feat: exponential backoff for rpc calls

ChaoticTempest:
Thought this was best to add next since CI builds are non-deterministic with sleeps and would fail sometimes. So all internal sleeps between each `call`/`view` function call are gone now. Also, won't likely have this for a while until near-api-rs is a thing in the middle, so might as well add it now and move it later.

`client.rs` handles most of the retry logic and also houses the logic for sending transactions

Also after this PR, completely addresses https://github.com/near/workspaces-rs/issues/7 by exposing mostly `anyhow::Result` for error handling, with the exception of one function left to be added by the spooning PR.

### DIFF
--- a/workspaces/src/rpc/api.rs
+++ b/workspaces/src/rpc/api.rs
@@ -1,4 +1,4 @@
-use super::client::client;
+use super::client;
 use super::tool;
 use super::types::{AccountInfo, NearBalance};
 
@@ -16,9 +16,7 @@ use near_jsonrpc_primitives::types::query::{QueryResponseKind, RpcQueryRequest};
 use near_primitives::borsh::BorshSerialize;
 use near_primitives::state_record::StateRecord;
 use near_primitives::transaction::SignedTransaction;
-use near_primitives::types::{
-    AccountId, Balance, BlockReference, Finality, FunctionArgs, Gas, StoreKey,
-};
+use near_primitives::types::{AccountId, Balance, Finality, FunctionArgs, Gas, StoreKey};
 use near_primitives::views::{FinalExecutionOutcomeView, QueryRequest};
 
 pub(crate) const NEAR_BASE: Balance = 1_000_000_000_000_000_000_000_000;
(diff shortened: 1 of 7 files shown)


workspaces/src/rpc/api.rs
```diff
@@ -48,20 +47,19 @@ impl From<FinalExecutionOutcomeView> for CallExecutionResult {
...
             },
         })
-        .await
-        .map_err(|err| err.to_string())?;
+        .await?;
 
     let account_view = match query_resp.kind {
         QueryResponseKind::ViewAccount(result) => result,
-        _ => return Err("Error call result".to_owned()),
+        _ => return Err(anyhow!("Error call result")),
```
austinabell, line 62: doesn't need to change since this can be more readable, but can use `bail!` for convenience https://docs.rs/anyhow/1.0.48/anyhow/macro.bail.html

workspaces/src/rpc/client.rs
```diff
@@ -0,0 +1,69 @@
...
+
+fn json_client() -> JsonRpcClient {
+    JsonRpcClient::connect(&rt_current_addr())
+}
+
+pub(crate) fn new() -> RetryClient {
+    RetryClient::new()
+}
+
+pub struct RetryClient;
```
austinabell, line 28: does this need to be pub and not crate scoped? I see it doesn't seem to be exported, but might be safe, just in case a refactor exposes this type?
```suggestion
pub(crate) struct RetryClient;
```

austinabell (approved):
LGTM, definitely an improvement and we can refine these APIs in a follow up to make sure it's as clean as possible

workspaces/src/rpc/api.rs
ChaoticTempest: ahh, didn't know about this macro. Will change it to `bail`. thanks!

austinabell (approved):
new change lgtm

Issue #21, opened 2021-11-10
Do not expose FinalExecutionOutcomeView


ChaoticTempest:
Currently, functions like https://github.com/near/workspaces-rs/blob/d12fe6bd887a37816de42a3ca34fcb7d77ec5705/workspaces/src/rpc/api.rs#L77 expose a `FinalExecutionOutcomeView` type as its return type. We don't need all that info and realistically only need things like gas consumed and whether the transaction succeeded w/ the proper error message if it failed

PR #22, opened 2021-11-17
Stop exposing `FinalExecutionOutcomeView` as the return type of many call-like functions

itegulov:
Resolves https://github.com/near/workspaces-rs/issues/21

The idea is to hide all the verbose information provided by `FinalExceutionOutcomeView` and its constituents. For now we are only interested in the outcome of the call along with the total gas burnt by the execution.


ChaoticTempest (approved):
Looking good to me

PR #23, opened 2021-11-25
feat: Rearrange API to rely less on macros

ChaoticTempest:
New API lets us create a worker instead of using macros:
```rust
use workspaces::prelude::*;

#[tokio::test]
async fn test_something() -> anyhow::Result<()> {
    let worker = workspaces::sandbox();

    let contract = worker.dev_deploy("path/to/contract.wasm").await?;
    let outcome = worker.call(&contract, "fn_name", serde_json::json! { /* args */ }, None).await?;
    println!("outcome: {:?}", outcome);

    Ok(())
}

```

This also removes all the macro-related code too. But could be added back later in the future as a more opinionated layer with a feature-gate.

Should be merged in after https://github.com/near/workspaces-rs/pull/20 merges.


workspaces/src/rpc/client.rs
```diff
@@ -2,41 +2,254 @@
...
+    async fn send_tx_and_retry(
+        &self,
+        signer: &InMemorySigner,
+        receiver_id: AccountId,
+        action: Action,
+    ) -> anyhow::Result<FinalExecutionOutcomeView> {
+        send_batch_tx_and_retry(self, signer, receiver_id, vec![action]).await
+    }
+
+    pub async fn call(
```
matklad, line 60: I think this is not visible outside of the crate. If that's the case, I suggest making it as `pub(crate)`. What we are developing is a public Rust API, so it is super-duper important to make sure that actually `pub` things have correct signatures, and for that, it's quite useful to tell at a glance whether the stuff is actually visible from outside of the crate or not.

workspaces/src/worker/impls.rs
```diff
@@ -0,0 +1,163 @@
...
+            )
+            .await
+            .map(Into::into)
+    }
+
+    pub async fn view(
+        &self,
+        contract_id: AccountId,
+        method_name: String,
+        args: FunctionArgs,
```
matklad, line 127: I feel strongly that we shouldn't be exposing types from `nearcore` in public API. Crates like `near-primitives` do not currently have public APIs and are woefully inadequate for use in public, stable APIs with backwards compatibility guarantees. 

Copy-pasting data definitions from nearcore is quite some amount of busy work. But without doing that, you'll be forced to break *clients* of workspaces-rs every time impl details of nearcore change.


In this specific case, additionally, it doesnt' make sense that call uses Vec<u8> while view uses `FunctionArgs`.

workspaces/src/network/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> anyhow::Result<()>
+    where
+        U: BorshSerialize + Send + Sync;
+}
+
+#[async_trait]
+impl<T> StatePatcher for T
+where
+    T: AllowStatePatching + NetworkClient + Send + Sync,
+{
```
austinabell, line 132: I feel like there should be a cleaner way to have this without exposing the AllowStatePatching trait. What you're trying to do here is just have a sealed trait impl, and you should be able to keep the AllowStatePatching trait private. I've taken a quick stab at this. This might not be the best alternative, but let me know your thoughts:

```
diff --git a/examples/src/spooning.rs b/examples/src/spooning.rs
index eff2653..20b059a 100644
--- a/examples/src/spooning.rs
+++ b/examples/src/spooning.rs
@@ -96,7 +96,11 @@ async fn main() -> anyhow::Result<()> {
 
     // Patch our testnet STATE into our local sandbox:
     worker
-        .patch_state(sandbox_contract.id(), "STATE".to_string(), &status_msg)
+        .patch_state(
+            sandbox_contract.id(),
+            "STATE".to_string(),
+            status_msg.try_to_vec().unwrap(),
+        )
         .await?;
 
     // Now grab the state to see that it has indeed been patched:
diff --git a/workspaces/src/network/mod.rs b/workspaces/src/network/mod.rs
index 30bcf02..21a734a 100644
--- a/workspaces/src/network/mod.rs
+++ b/workspaces/src/network/mod.rs
@@ -10,7 +10,6 @@ use async_trait::async_trait;
 
 use near_crypto::{InMemorySigner, KeyType, Signer};
 use near_jsonrpc_client::methods::sandbox_patch_state::RpcSandboxPatchStateRequest;
-use near_primitives::borsh::BorshSerialize;
 use near_primitives::state_record::StateRecord;
 use near_primitives::types::AccountId;
 
@@ -111,35 +110,31 @@ where
     }
 }
 
-pub trait AllowStatePatching {}
+pub(crate) mod sealed {
+    pub trait AllowStatePatching {}
+}
 
 #[async_trait]
-pub trait StatePatcher {
-    async fn patch_state<U>(
+pub trait StatePatcher: sealed::AllowStatePatching {
+    async fn patch_state(
         &self,
         contract_id: AccountId,
         key: String,
-        value: &U,
-    ) -> anyhow::Result<()>
-    where
-        U: BorshSerialize + Send + Sync;
+        value: Vec<u8>,
+    ) -> anyhow::Result<()>;
 }
 
 #[async_trait]
 impl<T> StatePatcher for T
 where
-    T: AllowStatePatching + NetworkClient + Send + Sync,
+    T: sealed::AllowStatePatching + NetworkClient + Send + Sync,
 {
-    async fn patch_state<U>(
+    async fn patch_state(
         &self,
         contract_id: AccountId,
         key: String,
-        value: &U,
-    ) -> anyhow::Result<()>
-    where
-        U: BorshSerialize + Send + Sync,
-    {
-        let value = U::try_to_vec(value).unwrap();
+        value: Vec<u8>,
+    ) -> anyhow::Result<()> {
         let state = StateRecord::Data {
             account_id: contract_id,
             data_key: key.into(),
diff --git a/workspaces/src/network/sandbox.rs b/workspaces/src/network/sandbox.rs
index 0e13426..6f2347b 100644
--- a/workspaces/src/network/sandbox.rs
+++ b/workspaces/src/network/sandbox.rs
@@ -9,7 +9,7 @@ use near_crypto::{InMemorySigner, Signer};
 use near_primitives::types::{AccountId, Balance};
 
 use super::{
-    Account, AllowDevAccountCreation, AllowStatePatching, CallExecution, Contract, NetworkClient,
+    Account, AllowDevAccountCreation, sealed, CallExecution, Contract, NetworkClient,
     NetworkInfo, TopLevelAccountCreator,
 };
 use crate::network::server::SandboxServer;
@@ -48,7 +48,7 @@ impl Sandbox {
     }
 }
 
-impl AllowStatePatching for Sandbox {}
+impl sealed::AllowStatePatching for Sandbox {}
 
 impl AllowDevAccountCreation for Sandbox {}
 
diff --git a/workspaces/src/prelude.rs b/workspaces/src/prelude.rs
index a499862..3f38e55 100644
--- a/workspaces/src/prelude.rs
+++ b/workspaces/src/prelude.rs
@@ -1,4 +1,4 @@
 pub use crate::network::{
-    AllowDevAccountCreation, AllowStatePatching, DevAccountDeployer, DevNetwork, Network,
+    AllowDevAccountCreation, DevAccountDeployer, DevNetwork, Network,
     StatePatcher, TopLevelAccountCreator,
 };
diff --git a/workspaces/src/worker/impls.rs b/workspaces/src/worker/impls.rs
index df803cb..a4cbac1 100644
--- a/workspaces/src/worker/impls.rs
+++ b/workspaces/src/worker/impls.rs
@@ -3,12 +3,11 @@ use std::path::Path;
 
 use async_trait::async_trait;
 use near_crypto::InMemorySigner;
-use near_primitives::borsh::BorshSerialize;
 use near_primitives::types::{AccountId, Balance, FunctionArgs, StoreKey};
 
 use crate::network::{
-    Account, AllowDevAccountCreation, CallExecution, CallExecutionDetails, Contract, NetworkClient,
-    NetworkInfo, StatePatcher, TopLevelAccountCreator,
+    sealed, Account, AllowDevAccountCreation, CallExecution, CallExecutionDetails, Contract,
+    NetworkClient, NetworkInfo, StatePatcher, TopLevelAccountCreator,
 };
 use crate::rpc::client::Client;
 use crate::worker::Worker;
@@ -74,20 +73,19 @@ where
     }
 }
 
+impl<T> sealed::AllowStatePatching for Worker<T> where T: sealed::AllowStatePatching {}
+
 #[async_trait]
 impl<T> StatePatcher for Worker<T>
 where
     T: StatePatcher + Send + Sync,
 {
-    async fn patch_state<U>(
+    async fn patch_state(
         &self,
         contract_id: AccountId,
         key: String,
-        value: &U,
-    ) -> anyhow::Result<()>
-    where
-        U: BorshSerialize + Send + Sync,
-    {
+        value: Vec<u8>,
+    ) -> anyhow::Result<()> {
         self.workspace.patch_state(contract_id, key, value).await
     }
 }
diff --git a/workspaces/tests/patch_state.rs b/workspaces/tests/patch_state.rs
index 4f1c513..49a6af8 100644
--- a/workspaces/tests/patch_state.rs
+++ b/workspaces/tests/patch_state.rs
@@ -72,7 +72,11 @@ async fn test_patch_state() -> anyhow::Result<()> {
     });
 
     worker
-        .patch_state(contract_id.clone(), "STATE".to_string(), &status_msg)
+        .patch_state(
+            contract_id.clone(),
+            "STATE".to_string(),
+            status_msg.try_to_vec().unwrap(),
+        )
         .await?;
 
     let result = worker

```

I also applied the change to Vec<u8> instead of being generic on borsh. This pattern can also be applied to the `AllowDevCreation` trait.

workspaces/src/network/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    async fn patch_state<U>(
+        &self,
+        contract_id: AccountId,
+        key: String,
+        value: &U,
+    ) -> anyhow::Result<()>
+    where
+        U: BorshSerialize + Send + Sync,
+    {
+        let value = U::try_to_vec(value).unwrap();
```
austinabell, line 142: Reason for unwrapping (if keeping this pattern)?

workspaces/src/network/testnet.rs
```diff
@@ -0,0 +1,109 @@
...
+        id: AccountId,
+        signer: InMemorySigner,
+        wasm: P,
+    ) -> anyhow::Result<CallExecution<Contract>> {
+        // TODO: async_compat/async version of File
+        let mut code = Vec::new();
+        File::open(wasm)?.read_to_end(&mut code)?;
+
+        let account = self.create_tla(id.clone(), signer.clone()).await?;
+        let account = Into::<anyhow::Result<_>>::into(account)?;
```
austinabell, line 73: Just noting there should probably be a more convenient API for converting into a result. Maybe we can add a `into_result` function on the `CallExecution`?

workspaces/src/prelude.rs
```diff
@@ -0,0 +1,4 @@
+pub use crate::network::{
+    AllowDevAccountCreation, AllowStatePatching, DevAccountDeployer, DevNetwork, Network,
```
austinabell, line 2: I very much doubt that the Allow.. traits need to be exposed. 

Also DevNetwork and Network, seems like there is no functionality tied to it? If needing to use, seems like this should just be imported normally. Maybe it's fine to keep, I'm just trying to make the prelude as slim as possible to avoid breaking changes if we ever want to remove anything

workspaces/src/worker/impls.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use crate::network::{
+    Account, AllowDevAccountCreation, CallExecution, CallExecutionDetails, Contract, NetworkClient,
+    NetworkInfo, StatePatcher, TopLevelAccountCreator,
+};
+use crate::rpc::client::Client;
+use crate::worker::Worker;
+
+unsafe impl<T> Send for Worker<T> where T: Send {}
+unsafe impl<T> Sync for Worker<T> where T: Sync {}
```
austinabell, line 17: We shouldn't have this here I don't think. What could be the reason for having this instead of auto impls?

workspaces/src/worker/impls.rs
```diff
@@ -0,0 +1,163 @@
...
+        &self,
+        id: AccountId,
+        signer: InMemorySigner,
+        wasm: P,
+    ) -> anyhow::Result<CallExecution<Contract>> {
+        self.workspace.create_tla_and_deploy(id, signer, wasm).await
+    }
+}
+
+impl<T> NetworkInfo for Worker<T>
```
austinabell, line 52: given all of the manual impls and this worker is just used to access the network, perhaps this is a use case for `Deref` and `DerefMut`. Definitely not the cleanest use case, but this feels like it could easily miss functionality or break as the lib gets built out.

wdyt @matklad ? Do you have an intuition that strongly favours yes or no here?

austinabell (commented):
A lot to cover so I definitely missed things and is not complete, but here are some comments!

matklad:
Unrelated question: why the thing is called `workspace`? If this is to mach terminology with other sdks (js, assembly script, etc), that that's fine. If this is rust-specific name, I storngly suggest reconsidering it -- it clashes badly with `cargo workspaces`.

README.md
```diff
@@ -1,38 +1,83 @@
...
+
+    let contract = worker.dev_deploy("path/to/file.wasm")
         .await
         .expect("could not dev-deploy contract");
 
-    let result = view(
+    let result = worker.view(
         contract_id,
         "function_name".to_string(),
         r#""some_arg": "some_value"".into(),
```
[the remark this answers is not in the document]
matklad: maybe a good use-case for `json!` macro.

workspaces/src/worker/impls.rs
matklad: My general intuition here is that all this type-and-trait-fu is not needed here. It seems simpler (for the user) if we stick with structs and inherent methods, and just return a runtime error for unsupported combinations of networks/requests:

```rust
pub struct Worker {
    rpc_addr: String,
    sandbox_process: Option<std::process::Child>,
}
```

austinabell:
> Unrelated question: why the thing is called `workspace`? If this is to mach terminology with other sdks (js, assembly script, etc), that that's fine. If this is rust-specific name, I storngly suggest reconsidering it -- it clashes badly with `cargo workspaces`.

Matching terminology across languages, yes. It was the name chosen unrelated to this Rust implementation specifically.

README.md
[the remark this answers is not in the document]
austinabell: ah, yeah, I guess I should have zoomed out and also made the comment that this is not valid JSON

examples/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2018"
 [dev-dependencies]
 anyhow = "1.0"
 tokio = { version = "1.10.0", features = ["full"] }
-workspaces = { version = "0.1.0", path = "../workspaces" }
+workspaces = { version = "0.2.0", path = "../workspaces" }
```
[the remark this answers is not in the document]
ChaoticTempest: fair enough, I was just upping the version just in case some people out there were already making use of the API as-is, but I did add a warning that things would change on the README. I don't feel strongly at all, so I can just keep it version-agnostic then

workspaces/src/network/account.rs
```diff
@@ -0,0 +1,53 @@
...
+        self.id.clone()
+    }
+
+    pub(crate) fn signer(&self) -> &InMemorySigner {
+        &self.signer
+    }
+}
+
+unsafe impl Sync for Account {}
+unsafe impl Send for Account {}
```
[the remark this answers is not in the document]
ChaoticTempest: Ahh, this was probably some of my hackery madness. Will remove it. Thanks for noticing!

workspaces/src/network/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn keystore_path(&self) -> std::path::PathBuf;
+
+    // TODO: change return type to Url instead of String
+    /// Rpc endpoint to point our client to
+    fn rpc_url(&self) -> String;
+
+    // TODO: not everything has a helper url. maybe make this optional or remove it into a seprate trait
+    /// The helper URL to create top level account out of for certain networks.
+    fn helper_url(&self) -> String;
+}
```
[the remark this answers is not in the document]
ChaoticTempest: Good point. I'll make it like the `NetworkClient` trait then

workspaces/src/network/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+        &self,
+        id: AccountId,
+        signer: InMemorySigner,
+    ) -> anyhow::Result<CallExecution<Account>>;
+
+    async fn create_tla_and_deploy<P: AsRef<Path> + Send + Sync>(
+        &self,
+        id: AccountId,
+        signer: InMemorySigner,
+        wasm: P,
```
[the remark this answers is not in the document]
ChaoticTempest: Gotcha, converted those over. And didn't add any helpers since adding `std::fs::read(filepath)` isn't that bad before each user defined deploy

workspaces/src/network/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    T: AllowStatePatching + NetworkClient + Send + Sync,
+{
+    async fn patch_state<U>(
+        &self,
+        contract_id: AccountId,
+        key: String,
+        value: &U,
+    ) -> anyhow::Result<()>
+    where
+        U: BorshSerialize + Send + Sync,
```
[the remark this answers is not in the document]
ChaoticTempest: Got rid of this to take in `Vec<u8>` instead

workspaces/src/rpc/client.rs
ChaoticTempest: I was thinking of giving this out to everyone to be able to make their own custom Networks later on (similarly w/ all the traits), but for now I'll make it all crate private just so we don't confuse anyone and see what people want out of it. And then maybe later expose it

workspaces/src/worker/impls.rs
ChaoticTempest: Ah, this was also one of my TODOs, but that fell through as I was rushing it for thanksgiving. This one adds quite a bit of types, so let me make a separate issue out of this and we can fix it up after this PR. Sounds good to you?

README.md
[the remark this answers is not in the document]
ChaoticTempest: I already use `json!` everywhere else. I guess I left it in since it was there for a while, and was too lazy to change. Thanks for catching it 👍

workspaces/src/network/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> anyhow::Result<CallExecution<Contract>>;
+}
+
+// NOTE: Not all networks/runtimes will have the ability to be able to do dev_deploy.
+// This trait acts as segmented boundary for only specific networks such as sandbox and testnet.
+pub trait AllowDevAccountCreation {}
+
+#[async_trait]
+pub trait DevAccountDeployer {
+    fn dev_generate(&self) -> (AccountId, InMemorySigner);
```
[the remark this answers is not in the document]
ChaoticTempest: I'll make the function `async` for now, but I'll add a separate issue to see which crate to make use of. Probably best to use something like `async-compat` since this library would now allow any runtime to be ran w/ it.

workspaces/src/network/mod.rs
ChaoticTempest: The pattern I was going for here was to allow people to have access to the `AllowStatePatching` trait so that if people wanted to implement their own network, they would just `impl AllowStatePatching for CustomNetwork` and get `StatePatcher` for free, otherwise they impl their own custom `StatePatcher` if they needed. I'm not sure if what I did was the best way to convey that functionality though. Let me know if there's a better way to do that or if we want to go for something simpler than this.

Also, I did clean up the trait to take in `Vec<u8>` instead of `T: BorshSerialize`.

workspaces/src/network/mod.rs
ChaoticTempest: Didn't keep this pattern, so resolving this one

workspaces/src/network/testnet.rs
ChaoticTempest: Sounds good with me

workspaces/src/prelude.rs
ChaoticTempest: True, I'll delete those. Definitely a lot of stuff I left there because it worked while I was hacking around with ideas. Much appreciated for doing such a thorough job of combing through my code like this

workspaces/src/worker/impls.rs
ChaoticTempest: I must have gotten rid of something that required it once upon a time. Thanks for pointing it out

workspaces/src/worker/impls.rs
ChaoticTempest: @matklad So, I could see us supporting custom networks in the future which is why there's all these different traits and such. `workspaces-js` already has similar stuff but under abstract classes instead, and I was trying to model off of that a little bit. 

Either ways, missing a trait impl isn't that bad here, since the compiler would just complain about it. If anything, allowing users to do `dev_deploy` on a concrete type/network that doesn't support it such as mainnet could be dangerous, because at that point, we're kinda relying on the devs to bail us out here. So the traits here would act as our separation of concern just in case something like that happens and we fall back on the compiler to help us here

ChaoticTempest:
@matklad 

in regards to the too many traits, I tried minimizing it as much as possible. I'm just afraid that later on, something will be missed and having these traits would prove useful in that certain functionality can only be achieved with certain networks, and are verifiable by the compiler itself. Like for mainnet, if we had `patch_state`, but the user already did some expensive stuff before hand all for it to fail on us at `patch_state`. That would be pretty infuriating dev experience. I'm trying to avoid that and have all of this be in checked during compile time. This gives us the freedom to not worry about such things later on. Also, certain things like `dev_deploy` is pretty prominent in the CLI so that's already a proven usage IMO.

> saw some places where we read/write stuff from/to disk

After moving some of the wasm filepath stuff out to user land instead, the only io operations left were saving keys. This one's pretty minor I think and we can just move it into another issue later

> it clashes badly with cargo workspaces

we somehow reserved `workspaces` name on crates.io though 😄. we also have `near-workspaces` too if we prefer that instead

workspaces/src/network/mod.rs
austinabell: hmm, yeah I suppose we want to expose the ability to configure a custom network, just seems excessive with the traits

ChaoticTempest: I don't know what the alternative would be though. Seems like this is as much as I can reduce it down to: as of now there's three traits in the prelude and two traits (`Network`, `DevNetwork`) we can directly import. I feel like those are all the essentials, and I've left some of the internal ones like `Allow*` to be exposed later

workspaces/src/worker/impls.rs
ChaoticTempest: addressed in https://github.com/near/workspaces-rs/pull/25

workspaces/src/network/server.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn start(&mut self) -> anyhow::Result<()> {
+        println!("Starting up sandbox at localhost:{}", self.rpc_port);
+        let home_dir = Sandbox::home_dir(self.rpc_port);
+
+        // Remove dir if it already exists:
+        let _ = std::fs::remove_dir_all(&home_dir);
+        near_sandbox_utils::init(&home_dir)?.wait()?;
+
+        let child = near_sandbox_utils::run(&home_dir, self.rpc_port, self.net_port)?;
+        println!("Started sandbox: pid={:?}", child.id());
```
austinabell, line 31: We probably want to switch these prints to some actual logging solution, but can tackle it in another issue which I will create after this review

austinabell (approved):
LGTM, definitely an improvement and in the direction I think we want to move to

ChaoticTempest:
@matklad no rush but wanted to see if there's any other concern before I merge this one in

matklad:
>no rush but wanted to see if there's any other concern before I merge this one in

Ah, sorry, didn't mean to block. There's definitely a rush -- this is a big architecture work, which blocks follow-ups. I feel we generally should try to merge such stuff faster. The only place where it makes to go *slower* is actually releasing thigs to the public -- that's when we can't take back our descisions. 

So, yeah, let's merge this! 

I still think though that having fewer traits and leaning onto runtime, rather than compile time, would be the right decision here, but tha's definitelly can be done in a follow up, and I need some time to lay my arguments down (I am not sure I am right on this one :) )

matklad:
Filed https://github.com/near/workspaces-rs/issues/31 to bikeshed about traits.

PR #25, opened 2021-12-02
chore: Upgrade jsonrpc-client & do not expose near_* crate types 

ChaoticTempest:
This PR should wait until https://github.com/near/workspaces-rs/pull/23 gets merged in first.

Just some chores and code cleaned up:
- No longer exposes `near-crypto` and `near-primitives` types to outside users since that could be dangerous if anything were to change with nearcore, plus `jsonrpc-client` relies heavily on specific versions of these types.
- `borsh` no longer exposed and removed since we're no longer relying on some of its traits for patching state.
- Upgraded `jsonrpc-client` to 0.1 since that is now stable and on crates.io.


workspaces/src/types.rs
```diff
@@ -0,0 +1,243 @@
...
+#[derive(Eq, Hash, Clone, Debug, PartialEq)]
+pub enum ParseErrorKind {
+    TooLong,
+    TooShort,
+    Invalid,
+}
+
+/// Signer that keeps secret key in memory.
+#[derive(Clone)]
+pub struct InMemorySigner {
```
matklad, line 132: Why do we need this type to be public in the first place? None of the examples use it. Can we just supply worker with secret keys? Something like 

```rust
impl<T> Worker<T> {
    pub fn register_account(&mut self, account_id: AccountId, key_pair: (PublicKey, SecretKey))
}
```

?

workspaces/src/types.rs
```diff
@@ -0,0 +1,243 @@
...
+        Self {
+            account_id: key_file.account_id,
+            public_key: key_file.public_key,
+            secret_key: key_file.secret_key,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct KeyFile {
```
matklad, line 203: Same here: do we need to expose this as an extra type in the API surface, or would something like 

```rust
impl Testnet {
    pub fn from_key_file(path: &Path) -> Self
}
```

would be enough?

workspaces/src/types.rs
```diff
@@ -0,0 +1,243 @@
...
+            #[cfg(target_os = "macos")]
+            perm.set_mode(u32::from(libc::S_IWUSR | libc::S_IRUSR));
+            #[cfg(not(target_os = "macos"))]
+            perm.set_mode(libc::S_IWUSR | libc::S_IRUSR);
+
+            file.set_permissions(perm)
+                .expect("Failed to set permissions for a key file.");
+        }
+
+        let str = serde_json::to_string_pretty(self).expect("Error serializing the key file.");
```
austinabell, line 230: prob want to avoid using `str` as the keyname since it's part of Rust prelude. Didn't even know this let you

workspaces/src/types.rs
```diff
@@ -0,0 +1,243 @@
...
+                ));
+            }
+            last_char_is_separator = current_char_is_separator;
+        }
+
+        (!last_char_is_separator)
+            .then(|| ())
+            .ok_or_else(|| ParseAccountError(ParseErrorKind::Invalid, account_id.to_string()))
+    }
+}
```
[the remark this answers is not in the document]
ChaoticTempest: It exposes a near_primitive error though. What's the general strategy for exposing those? I think I could map that to the version here instead of duplicating all the validate logic

workspaces/src/types.rs
```diff
@@ -0,0 +1,243 @@
...
+use near_crypto::key_conversion::convert_secret_key;
+use near_crypto::vrf::{Proof, Value};
+use near_crypto::Signature;
+use near_primitives::account::id::{MAX_ACCOUNT_ID_LEN, MIN_ACCOUNT_ID_LEN};
+
+pub(crate) use near_crypto::{KeyType, PublicKey, SecretKey, Signer};
+use serde::{Deserialize, Serialize};
+
+#[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub struct AccountId(Box<str>);
```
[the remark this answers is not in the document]
ChaoticTempest: as in make a `newtype` out of the primitive one? If so yeah, I think that works

workspaces/src/types.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+
+impl From<AccountId> for near_primitives::types::AccountId {
+    fn from(id: AccountId) -> Self {
+        id.0.into_string()
+            .try_into()
+            .expect("Should not fail since this is already validated before")
+    }
+}
```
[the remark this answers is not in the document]
ChaoticTempest: Pretty much, but now that you pointed it out, that assumption could be wrong when we go to update the dependencies. In that case, yeah it'd be better to use TryFrom then. I'll switch it over

workspaces/src/types.rs
ChaoticTempest: It's because `create_tla` and `create_tla_and_deploy` functions require it. But I think we can just switch those to take in key_pair instead then

workspaces/src/types.rs
ChaoticTempest: Ahh yeah, this seems to be just a type we don't need to expose at all. Could just be `pub(crate)` or could just eliminate it completely and move all the functions into the impl of `Signer` and `InMemorySigner` where it gets used

workspaces/src/types.rs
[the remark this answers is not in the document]
matklad: `AccountId` specifically might be fine to expose. I got the feeling that we might publish 1.0 of it relatively soon cc @miraclx

workspaces/src/types.rs
matklad: Things I wish I didn't know part 92. 

Rust has namespaces, so it's OK to have values and types named the same: 

```rust
type X = ();
const X: X = ();
```

**but** the primitive types are doubly cursed, and have special-casing in the compiler. The `str` type and `std::str` module both live in the type namespace, but compiler still is able to disambiguate between them: 

```rust
use std::str;

fn main() {
    let assoc_fn = str::as_bytes;
    let fn_in_module = str::from_utf8;
}
```

workspaces/src/types.rs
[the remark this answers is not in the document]
miraclx: Yes indeed! `near-account-id 0.11.0` would be released later this week, sporting all the refactoring that's gone into it the past week. Also, while `near-account-id` is exposed from `near-primitives`, it's also its own [crate](https://docs.rs/near-account-id). So it's fine to expose.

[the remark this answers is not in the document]
matklad: Well, I'd say that 0.11 is not super great to expose -- it implies that 0.12 would be released some time in the future, and if we expose it publicly here, that means that we'd have to publish a major as well, and *that* means that all the people who write tests for contracts need to upgrade to the new major version. When we release 1.0, we should make sure that everything we expose is also `1.0`. 

I guess let's do this -- expose AccountId directly for now while iterating in 0.1 ourselves. Once we publish 1.0, either wrap it in our own type or (preferred) pressure miraclx into releasing 1.0 as well :)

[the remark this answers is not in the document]
austinabell: To be clear, my suggestion was to keep the method as is, just re-use the internal impl and convert the error type to our own. Maybe I'm missing something so I will play around with this when I get to my computer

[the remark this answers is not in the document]
ChaoticTempest: hmm, I don't have an opinion on this, so I'll just get rid of it for now and expose the near_primitive one. Let me know @austinabell if you feel differently

workspaces/src/types.rs
ChaoticTempest: Actually, there's a couple functions that would require it too, like `delete_account` and `transfer_near` that I think would be more intuitive if it had signer over a key pair.

matklad: TBH, I feel that these functions' signatures are less than ideal.

```rust
    pub(crate) async fn transfer_near(
        &self,
        signer: &InMemorySigner,
        receiver_id: AccountId,
        amount_yocto: Balance,
    ) -> anyhow::Result<FinalExecutionOutcomeView> {
```

From the user's POV, "signer" doesn't make sense. As a user, I want to transfer from one account id to another account id, so I'd expect `fn transfer(&self, from: AccountId, to: AccountId, amount: Ballance)`. The `Signer`, and especially `InMemory` bit, feel like impl details that leak into the interface. Can we hide this from the user completely?  Would something like this work?

```rust
fn test_transfer(w: Worker) {
   let alice: Account = w.create_account("alice");
   let bob: Account = w.create_account("bob");

   alice.transfer_near(bob.id(), 9001);
}
```

To put in other way, why to we need to expose `Signer` from `dev_generate` in the first place? 

```
async fn dev_generate(&self) -> (AccountId, InMemorySigner) {
```

Can we return an `Account` instead, or stash the signer in some kind of a hash map?

workspaces/src/types.rs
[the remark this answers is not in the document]
matklad: >and expose the near_primitive one

I belive you can use one from crates.io https://crates.io/crates/near-account-id. You might need the corresponding other crates from crates as well: 

* https://crates.io/crates/near-crypto
* https://crates.io/crates/near-primitives


Doing `pub use near_account_id::AccountId` is better than `pub use near_primitives::type::AccountId`. Although these are the same type right now, `near_primitives` does not guarantee that that'll be the case in the future as well.

workspaces/src/types.rs
ChaoticTempest: > why to we need to expose Signer from dev_generate in the first place?

Originally, just a helper to let us sign transactions more easily, but I changed it to return `SecretKey` instead now since that was good enough to create a signer from.

> Can we return an Account instead, or stash the signer in some kind of a hash map?

Yeah, I was going to add some convenience functions to `Account`s in a future PR. Already have one for `call` here: https://github.com/near/workspaces-rs/pull/33. We could also do what near-api-js does and have a global keystore associated to an account-id, but that might be a bit hard to track when keys are being registered. That was at least my experience when I was looking through the code. @volovyk-s what do you think about adding a `KeyStore`-like object to register accounts and tracks keys?

workspaces/src/types.rs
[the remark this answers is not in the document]
ChaoticTempest: Got it, switched it over to `near-account-id` instead

workspaces/Cargo.toml
```diff
@@ -19,13 +19,15 @@ tokio = { version = "1", features = ["full"] }
 tokio-retry = "0.3"
 url = { version = "2.2.2", features = ["serde"] }
 
-near-crypto = { git = "https://github.com/near/nearcore", rev = "88a9e74d4b99867e417fe199cbd74559fd4fc349" }
-near-primitives = { git = "https://github.com/near/nearcore", rev="88a9e74d4b99867e417fe199cbd74559fd4fc349" }
-near-jsonrpc-primitives = { git = "https://github.com/near/nearcore", rev="88a9e74d4b99867e417fe199cbd74559fd4fc349" }
-
+near-account-id = "0.5"
+near-crypto = "0.5"
+near-primitives = "0.5"
```
austinabell, line 24: there is `0.10` version of this https://crates.io/crates/near-primitives/versions and I'm not sure the difference but we use this in the SDK so if we want to de-dup deps we might want to make sure it is synced.

workspaces/src/rpc/tool.rs
```diff
@@ -47,11 +50,44 @@ pub(crate) async fn url_create_account(
...
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::prelude::PermissionsExt;
+        let mut perm = file
+            .metadata()
+            .expect("Failed to retrieve key file metadata.")
+            .permissions();
+
+        #[cfg(target_os = "macos")]
```
austinabell, line 73: will this even be hit when in the block of `cfg(unix)` or does that only define what is compiling the code?

austinabell (approved):
LGTM generally, but curious about these two things

workspaces/Cargo.toml
ChaoticTempest: `near-jsonrpc-client` only accepts this specific version of those primitives, which is why I couldn't up them unless we want to create bridges from one version to the next

workspaces/src/rpc/tool.rs
ChaoticTempest: macos is apart of the `unix` target_family so it should be hit. I noticed this because the macos tests were failing to compile for this case

austinabell: oh wow I completely blanked on that question, disregard :D

workspaces/src/types.rs
volovyk-s: @ChaoticTempest KeyStore is a good pattern, I think we can use it here. Ideally, it should be a part of the unexisting `near-api-rs` library :/
You can check the pattern in `near-api-js` for now.

ChaoticTempest: Ok sounds good then. Let's just go with the convenience functions in a later PR then

workspaces/src/rpc/tool.rs
ChaoticTempest: ahh no worries

Issue #31, opened 2021-12-06
Traits or no traits


matklad:
There are two alternative ways we can design the overall shape API.

One is to make the Worker type to be parametrized by network type. So things like `Sandbox` and `TestNet` are different types. The public API would look roughly like this:

```rust
 /// Worker holds an actual network, a distinct type
 pub struct Worker<T> { }

/// Methods which can be specific to some network are abstracted away in a
 /// trait.
 #<span class="error">[async_trait]</span>
 pub trait TopLevelAccountCreator 

 { async fn create_tla(&self, id: AccountId, signer: InMemorySigner) -> Result<CallExecution<Account>>; async fn create_tla_and_deploy(&self, id: AccountId, signer: InMemorySigner, wasm: Vec<u8>) -> Result<CallExecution<Contract>>; } 

/// Worker has extra methods via trait if the underlying network supports it,
 /// checked at compile time
 #<span class="error">[async_trait]</span>
 impl<T> TopLevelAccountCreator for Worker<T>
 where
 T: TopLevelAccountCreator + Send + Sync,

 { // Delegates } 

/// Methods available on all of the networks are inherent.
 impl<T> Worker<T>
 where
 T: NetworkClient,
 {
 pub async fn call(&self, contract: &Contract, method: String, args: Vec<u8>, deposit: Option<Balance>) -> Result<CallExecutionDetails> { }
 pub async fn view(&self, contract_id: AccountId, method_name: String, args: FunctionArgs) -> Result<serde_json::Value> { }
 }

pub struct Sandbox { }
 impl NetworkClient for Sandbox {}
 #<span class="error">[async_trait]</span>
 impl TopLevelAccountCreator for Sandbox {}

pub struct Testnet { }
 impl NetworkClient for Sandbox {}
 // NB: this impl is **not** provided.
 // #<span class="error">[async_trait]</span>
 // impl TopLevelAccountCreator for Sandbox {}
 ```

The usage would look like this: 

```rust
 use workspaces::

 {Worker, Sandbox, TopLevelAccountCreator} 

;

fn main() 

 { let w: Worker<Sandbox> = Worker::new(); .... } 

fn generic_usage<T: TopLevelAccountCreator + NetworkClient>(w: Worker<T>) 

 { ... } 

```

An alternative is to **not** make type-level distinction between the types of network. The public API would then look like so: 

```rust
 /// Worker holds a JSON rpc clinet, the same type for all networks.
 pub struct Worker{ }

impl Worker {
[...]

Issue #28, opened 2021-12-03
Add convenience functions for working with Account and Contract types


ChaoticTempest:
Instead of passing contract/account into `worker.call()`, we can add some wrappers around it within account/contract itself.

For instance, `account.create_account` or `contract.view` would be more ergonomic

Non-exhaustive list of helpers:

*   account
*   <span class="error">[x]</span> create_account (sub-account)
*   [ ] create_and_deploy
*   <span class="error">[x]</span> balance / available_balance
*   <span class="error">[x]</span> delete (deletes the account, and sends balance to beneficiary)

*   contract
*   <span class="error">[x]</span> call
*   <span class="error">[x]</span> view
*   <span class="error">[x]</span> view_code
*   [ ] view_state
*   [ ] patch_state (sandbox only)

shariffdev:
@frol I worked on this one and it can be closed.

PR #33, opened 2021-12-08
feat: convenience functions for `Contract` and `Account`

ChaoticTempest:
requires https://github.com/near/workspaces-rs/pull/25 to be merged in first.

addresses one of the bullet points in https://github.com/near/workspaces-rs/issues/28

### Account/Contract Convenience Functions
```rust
let worker = workspaces::sandbox();
let contract = worker.dev_deploy(...).await?;
let account = worker.dev_create().await?;

account.call(&worker, contract_id, "function".into())
   .with_args( ... )
   .with_gas( /*optional: defaults to an internal constant */ )
   .with_deposit( /* optional: defaults to 0 */)
   .transact()
   .await?;
```

Can call directly into the contract using the contract's signer to sign the transaction itself. Does not require to specify the contract_id 
```rust
contract.call(&worker, "function".into())
   .with_args( ... )
   .with_gas( /*optional: defaults to an internal constant */ )
   .with_deposit( /* optional: defaults to 0 */)
   .transact()
   .await?;
```

`account.transfer_near`, `account.delete_account`, `account.deploy` and `contract.delete_contract` now available


workspaces/src/network/account.rs
```diff
@@ -17,6 +20,44 @@ impl Account {
     pub(crate) fn signer(&self) -> &InMemorySigner {
         &self.signer
     }
+
+    /// Call a contract on the network specified within `worker`, and return
+    /// a Builder object that we will make use to populate the rest of the
+    /// call details.
+    pub fn call<'a, T: Network>(
+        &self,
+        worker: &'a Worker<T>,
```
matklad, line 29: I wonder if the `Account` should just store an `Arc<Worker>` internally, so that the user doesn't have to pass it everywhere?

willemneal:
@ChaoticTempest shouldn't it be `account_id` to stay consistent with the sdk?

workspaces/src/network/account.rs
ChaoticTempest: That sounds like a good idea to me

ChaoticTempest:
@TrevorJTClarke Equivalent in mainnet would be `worker.create_tla`/`worker.create_tla_and_deploy`. I haven't exposed sub-account creation yet, but let me add it to this PR too since it's another convenience function for account.

Creating a sub account should look like this:
```rust
let subaccount = account.create_account("subaccount")
    .with_initial_balance(parse_near!("1 N")); // optional: defaults to 0.1 Near. Enough to hold storage
    .with_key(secret_key)  // optional: if not specified, generate it
    .transact()
    .await?;

println!("account id: {:?}", account.id());  // should yield "dev-{NUM}"
println!("subaccount id: {:?}", subaccount.id());   // should yield "subaccount.dev-{NUM}"
```

@willemneal 
> @ChaoticTempest shouldn't it be `account_id` to stay consistent with the sdk?

You mean how we reference `.id()` instead of `.account_id()`? I think that's only apart of `near-sdk-sim` so we don't need to make it consistent just quite.

workspaces/src/network/account.rs
ChaoticTempest: Won't be able to have this in this PR unfortunately since we'd require `Account` to take a type parameter. Let's just wait until we resolve https://github.com/near/workspaces-rs/issues/31 and purely just get the `Worker` type without the type parameter

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,188 @@ impl Contract {
...
+
+    pub async fn transact(self) -> anyhow::Result<CallExecutionDetails> {
+        self.worker
+            .client()
+            .call(
+                &self.signer,
+                self.contract_id,
+                self.function,
+                self.args
+                    .expect("required `with_args` to be specified as apart of Call"),
```
austinabell, line 201: A call function doesn't need to use params, does it? Why not just default to empty vec if it isn't required (I'm not 100% sure it isn't required, but I can't see why it would be)

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,188 @@ impl Contract {
...
+    }
+
+    pub async fn view(self) -> anyhow::Result<Vec<u8>> {
+        self.worker
+            .client()
+            .view(
+                self.contract_id,
+                self.function,
+                self.args
+                    .expect("required `with_args` to be specified as apart of View"),
```
austinabell, line 216: same here

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,188 @@ impl Contract {
...
+
+impl<'a, T> CreateAccountBuilder<'a, T>
+where
+    T: Network,
+{
+    fn new(
+        worker: &'a Worker<T>,
+        signer: InMemorySigner,
+        parent_id: AccountId,
+        new_account_id: String,
```
austinabell, line 241: I'm a little confused on why we are forcing the account to be built as a subaccount. If this is specifically intended for that flow, perhaps we should rename to indicate this?

workspaces/src/rpc/client.rs
```diff
@@ -96,7 +96,7 @@ impl Client {
         };
 
         let result = std::str::from_utf8(&call_result)?;
-        Ok(serde_json::from_str(result)?)
+        Ok(result.into())
```
austinabell, line 99: This should probably be bytes. View calls can return anything, and esp when using borsh returns this will be unusable

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,188 @@ impl Contract {
...
+            signer,
+            contract_id,
+            function,
+            args: None,
+            deposit: None,
+            gas: None,
+        }
+    }
+
+    pub fn with_args(mut self, args: Vec<u8>) -> Self {
```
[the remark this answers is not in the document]
ChaoticTempest: That would be pretty useful. I'll add it, thanks for the suggestion

workspaces/src/network/account.rs
ChaoticTempest: So actually, empty vec fails for (certain or maybe all) calls. We'd at least need empty brackets such as `serde_json::json!({})`, but not sure what the equivalent borsh would be though. I think we can just default to `{}` for now and force people to do `with_args_borsh` if they expect their args to be borsh serialized

workspaces/src/network/account.rs
ChaoticTempest: I'll just rename it to be `create_subaccount` then. This was initially to be consistent with `workspaces-js/Account`, but better to be explicit here

workspaces/src/network/account.rs
austinabell: Wait, why is this a requirement? Is it with the API? There should not be this requirement. Args should be optional

ChaoticTempest: Actually there's an interesting case with this. Empty `Vec::new()` fails for functions taking all `Option`s, so we're required to supply `{}` here instead. Not sure why that is the case

ChaoticTempest: Also, just tested it purely with a function that takes no arguments, and it took both `Vec::new()` and `json!({})` without a problem. I think it's just how json rpc is working under the hood. We can take a closer look another time, but let's just go with a default of `json!({})` so it fits both cases. WDYT?

workspaces/src/network/account.rs
austinabell: is the fact that it's a subaccount obfuscated completely from the API? The only issue I see is just that it may be doing something differently than a developer may expect, which is a bit of a foot gun. Maybe we should consider renaming on js side for consistency as well if we want to make this change?

workspaces/src/network/account.rs
```diff
@@ -169,14 +169,24 @@ impl<'a, T: Network> CallBuilder<'a, T> {
...
         }
     }
 
     pub fn with_args(mut self, args: Vec<u8>) -> Self {
-        self.args = Some(args);
+        self.args = args;
+        self
+    }
+
+    pub fn with_args_json(mut self, args: serde_json::Value) -> Self {
```
austinabell, line 183: prob more convenient to just have this generic on serde serialize, to avoid a user to convert to a serde_json::Value and avoid exposing this detail to the user?

workspaces/src/network/account.rs
ChaoticTempest: > is the fact that it's a subaccount obfuscated completely from the API?

Yup, it's pretty much that. When we go to call into account creation, there's no flag to say, "hey let's go make a subaccount". Instead, it's just based on account id we give it such as "A" or "B.A"

> Maybe we should consider renaming on js side for consistency as well if we want to make this change?

The js side is pretty flexible (maybe too flexible). The `create_account` on that side allows you to create both top level and subaccounts regardless of what `account` context is present. Changing it on the js side would probably require a good amount of rework. On rust side, I've just had it be explicit in the functionality. If you want to create a TLA account you call `worker.create_tla`, and subaccounts with `account.create_subaccount`

workspaces/src/rpc/client.rs
ChaoticTempest: Changed this to use `ViewResultDetails` instead

workspaces/src/network/account.rs
austinabell: okay, I think I might have to look through code when I get a chance to understand this difference, thanks for giving context!

workspaces/src/network/account.rs
ChaoticTempest: sounds good to me, and added

workspaces/src/network/account.rs
ChaoticTempest: it's fine, I just opened one up here: https://github.com/near/workspaces-rs/issues/36

workspaces/src/network/account.rs
ChaoticTempest: Added an issue for this: https://github.com/near/workspaces-rs/issues/37

workspaces/src/network/result.rs
```diff
@@ -53,3 +53,32 @@ impl From<FinalExecutionOutcomeView> for CallExecutionDetails {
         }
     }
 }
+
+/// The result from a call into a View function. This contains the contents or
+/// the results from the view function call itself. The consumer of this object
+/// can choose how to deserialize its contents.
+pub struct ViewResultDetails {
```
matklad, line 60: `#[non_exhaustive]`

workspaces/src/network/result.rs
```diff
@@ -53,3 +53,32 @@ impl From<FinalExecutionOutcomeView> for CallExecutionDetails {
...
+    /// Our result from our call into a view function.
+    pub result: Vec<u8>,
+    /// Logs generated from the view function.
+    pub logs: Vec<String>,
+}
+
+impl ViewResultDetails {
+    pub fn try_serde_deser<'a, T: serde::Deserialize<'a>>(&'a self) -> anyhow::Result<T> {
```
matklad, line 68:
```suggestion
    pub fn json<T: serde::DeserializeOwned>(&self) -> anyhow::Result<T> {
```

workspaces/src/network/result.rs
```diff
@@ -53,3 +53,32 @@ impl From<FinalExecutionOutcomeView> for CallExecutionDetails {
...
+        serde_json::from_slice(&self.result).map_err(Into::into)
+    }
+
+    pub fn try_borsh_deser<T: borsh::BorshDeserialize>(&self) -> anyhow::Result<T> {
```
matklad, line 72:
```suggestion
    pub fn borsh<T: borsh::BorshDeserialize>(&self) -> anyhow::Result<T> {
```

workspaces/tests/create_account.rs
```diff
@@ -0,0 +1,20 @@
+use workspaces::prelude::*;
+
+#[tokio::test]
+async fn test_subaccount_creation() -> anyhow::Result<()> {
+    let worker = workspaces::sandbox();
+    let account = worker.dev_create().await?;
```
matklad, line 6: `dev_create` seems like less than ideal name. Dev create what? Maybe `worker.test_account()` or `worker.new_test_account()`? `worker.test_account()` which always returns *the same* account sounds like a potentially good idea.

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,191 @@ impl Contract {
...
+        self.deposit = Some(deposit);
+        self
+    }
+
+    pub fn with_gas(mut self, gas: u64) -> Self {
+        self.gas = Some(gas);
+        self
+    }
+
+    pub async fn transact(self) -> anyhow::Result<CallExecutionDetails> {
```
matklad, line 203: Somewhat awkward name, not sure what would be better `.call` perhaps?

workspaces/src/rpc/client.rs
```diff
@@ -72,13 +73,12 @@ impl Client {
         .await
     }
 
-    // TODO: return a type T instead of Value
```
matklad, line 75: Unrelated to the PR, but in the `call` method, I'd made all parameters required, and left defaults to the builder. It's better to set defaults once, at the outer layer.

matklad (approved):
LGTM overall! 

There's some awkwardness in API here and there, which is totally fine for the current state of iteration!

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,191 @@ impl Contract {
...
+        self
+    }
+
+    pub fn with_args_json<U: serde::Serialize>(mut self, args: U) -> Self {
+        self.args = serde_json::to_vec(&args).unwrap();
+        self
+    }
+
+    pub fn with_args_borsh<U: borsh::BorshSerialize>(mut self, args: U) -> Self {
+        self.args = args.try_to_vec().unwrap();
```
austinabell, line 189: prob best for these to be functions that return result rather than panicking

workspaces/src/network/account.rs
austinabell: I think the rationale here, and just my assumption, is just that `transact` makes it more clear that it's signing and sending a transaction rather than just calling a function in a view context, but this might be intuitive for a developer anyway.

workspaces/src/network/result.rs
austinabell: Why use DeserializeOwned when there can be references to the buffer in the deserialization? Am I missing something where there are no cases of this with json? But yes, I think naming for json to be more explicit would be best

matklad: Mostly "because reqwest does this", but I can see several benefits here:

* Simplicity for the user
* Pit of success -- non-owned deserialize is a very niche thing, and I don't think it can ever be useful here. But I can see users spending extra mental cycles to make borrowed deserialization work simply because that's available. 
* THere's also angle that DeserializeOwned is more flexible for us, and would allow, eg, to receive bytes into a non-contiguous buffer. But that's a bad point, as we expose `Vec<u8>` anyway, and we *should* expose it for simplicity.

matklad: No strong opinion though: `Deserialize<'a>` would be ok. I do feel much stronger about method's name though.

austinabell: My guess for reqwest doing this is maybe because they are deserializing from a stream rather than a buffer? 

Yeah I think it's definitely niche, but as to your last point: I also just don't think it will ever make sense in this example for part of the response to be some stream that we keep open until used and disallow using more than once.

I guess the argument against this could be that a user might not know they can use `DeserializeOwned` in place of this if they are creating some convenience function with generics, but this also seems like a niche case.

Might be bikeshedding at this point, so I'll just make it clear I'm fine with both ways, just figured I'd think through

workspaces/src/network/account.rs
```diff
@@ -17,6 +24,60 @@ impl Account {
...
     }
+
+    /// Call a contract on the network specified within `worker`, and return
+    /// a Builder object that we will make use to populate the rest of the
+    /// call details.
+    pub fn call<'a, T: Network>(
+        &self,
+        worker: &'a Worker<T>,
+        contract_id: AccountId,
+        function: String,
```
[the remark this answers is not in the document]
ChaoticTempest: I'll change the function signature since that would also be easier on the user to specify static strings, but underneath, it still needs to be an owned string due to how `near_primitives::transaction::FunctionCallAction` requires an owned string as an argument

workspaces/src/network/account.rs
```diff
@@ -17,6 +24,60 @@ impl Account {
...
+    }
+
+    /// Transfer near to an account specified by `receiver_id` with the amount
+    /// specified by `amount_yocto`. Returns the execution details of this
+    /// transaction
+    pub async fn transfer_near<T: Network>(
+        &self,
+        worker: &Worker<T>,
+        receiver_id: AccountId,
+        amount_yocto: Balance,
```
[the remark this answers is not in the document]
ChaoticTempest: `Balance` is the type that gets used everywhere for this even such as `api-js`. It make sense for it to be balance since you are transferring `Balance`s between accounts. The name `Near` is already a bit bloated in the ecosystem is the thing. How about I just have it be `amount: Balance`? - since `Balance` is by default yocto and the expectation would be that a user use one of the utility functions such as `pase_near!("10 N")`

workspaces/src/network/account.rs
```diff
@@ -40,7 +101,191 @@ impl Contract {
         &self.account.id
     }
 
+    pub fn as_account(&self) -> &Account {
```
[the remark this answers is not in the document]
ChaoticTempest: The contract's account is not necessarily the owner though, since the behavior for an owner can belong to the parent account or the account that created this `Contract`/`Account`

workspaces/tests/create_account.rs
ChaoticTempest: so I wanted to tie this with `dev_deploy` since they're very similar except `dev_deploy` also deploys a contract. How about `dev_create_account()`? Returning the same account will be problematic for testnet since that could cause some non-determinism if every testnet test used the same account and modified the accounts state

workspaces/src/network/result.rs
ChaoticTempest: No strong opinions here either, but let's just go with simplicity for now and have `DeserializeOwned`. I'm using it already in the ref-finance example/PR for `CallExecutionDetails` so this will make it consistent with that too

workspaces/src/network/account.rs
```diff
@@ -40,7 +102,191 @@ impl Contract {
...
+    ) -> CallBuilder<'a, T> {
+        self.account.call(worker, self.id().clone(), function)
+    }
+
+    /// Call a view function into the current contract. Returns a result that
+    /// yields a JSON string object.
+    pub async fn view<T: Network>(
+        &self,
+        worker: &Worker<T>,
+        function: String,
```
austinabell, line 133: this is asymmetric with the signature above, we should probably do the same here.

austinabell (commented):
One thing I will note is that this is missing a lot of docs, we can always make a pass through before releasing to make sure it's all there, but might be better to do some now when you have the context

workspaces/src/network/account.rs
```diff
@@ -79,6 +79,22 @@ impl Account {
             new_account_id,
         )
     }
+
+    pub async fn deploy<T: Network>(
```
TrevorJTClarke, line 83: @ChaoticTempest HOW IN THE WORLD ARE YOU THIS FAST

workspaces/src/network/result.rs
```diff
@@ -53,3 +55,33 @@ impl From<FinalExecutionOutcomeView> for CallExecutionDetails {
...
+    /// Logs generated from the view function.
+    pub logs: Vec<String>,
+}
+
+impl ViewResultDetails {
+    pub fn json<T: serde::de::DeserializeOwned>(&self) -> anyhow::Result<T> {
+        serde_json::from_slice(&self.result).map_err(Into::into)
+    }
+
+    pub fn borsh<T: borsh::BorshDeserialize>(&self) -> anyhow::Result<T> {
```
austinabell, line 75: I think docs on these would be very useful. It's very unclear by the names `json` and `borsh` that this is trying to deserialize the result imo.

austinabell (approved):
LGTM, definitely an improvement. The only note I'd make is that the examples it might be more clear if the account signing the transactions are an account other than the contract itself, since this is an uncommon pattern in practice

workspaces/src/network/account.rs
```diff
@@ -17,6 +25,76 @@ impl Account {
...
+            self.signer.clone(),
+            self.id().clone(),
+            new_account_id,
+        )
+    }
+
+    pub async fn deploy<T: Network>(
+        self,
+        worker: &Worker<T>,
+        wasm: Vec<u8>,
```
austinabell, line 86: I wonder if, like the other conversion from `String` to `&str`, this should also be done because in the future we can assume that we don't actually require an owned vec to be able to serialize and send over the wire?

ChaoticTempest: That's the thing about a lot of the `near_primitives`/rpc, a lot of it requires owned types. Like `DeployContractAction` takes in a `Vec<u8>` so I can't do much here unless we change the deps

workspaces/src/network/account.rs
ChaoticTempest: ¯\_(ツ)_/¯

workspaces/src/network/account.rs
matklad: Important meta note: whatever our *private* deps do internally should not have any bearing on the public API we expose. 

The RPC lib we use can use `&[u8]` or `Vec<u8>`, or `Bytes`, but that shouldn't change the public types. We should be able to refactor implementation from `Vec<u8>` to `Bytes` and back, if we like to, without making our user's change the code. 

In this sense, yeah, it seems that `wasm: &[u8]` would be a better signature here -- nothing in the problem fundamentally requires owned vec. I can also imagine the following idiom for sandbox testing: 

```
let CODE: &'static [u8] = include_bytes!("./target/wasm32-unknown-unknow/contrct.wasm");

#[test]
fn foo() {}
```

matklad: Well, context is the king: in some cases, you want the abstraction to be inherently leaky to have, eg, maximal performance. What we are doing here is essentially a scripting layer, so we have a lot of freedom in how we organize internals, and have a lot of responsibility in making the public API most natural for the user.

workspaces/tests/create_account.rs
matklad: >Returning the same account will be problematic for testnet since that could cause some non-determinism if every testnet test used the same account

I mean that "the same" is scoped to the `worker` -- basically, "create a new dev account" is something you use rarely, and for most tests, which need just a single account, you use "get the test account", which is created on the first access. 

Not sure whether this is a good idea...

Thinking more about this, it's probably a good idea -- for the first release we should aim for orthogonal API, and we can add shortcuts later.

matklad: Potentially horrible idea this is async function, so the user already has to type `.await`. So what we *can* do is

```rust
let account_without_code = worker.dev_create_account().await?;
let account_with_code = worker.dev_create_account().with_code(&wasm).await?;
```

This reads pretty nice at the call-site, but would be quite mind bending in the rustdoc. I wonder if there's some precedent somewhere for such async builders?

workspaces/src/network/account.rs
ChaoticTempest: hmm, let's just do `wasm: AsRef<[u8]>` then so we can take in both `Vec<u8>` and `&[u8]`

workspaces/tests/create_account.rs
ChaoticTempest: > I mean that "the same" is scoped to the worker -- basically, "create a new dev account" is something you use rarely, and for most tests, which need just a single account, you use "get the test account", which is created on the first access.

That pretty much is the `root_account` for Sandbox which covers most of those cases. For Testnet we can emulate this by doing what you describe. 

> Thinking more about this, it's probably a good idea -- for the first release we should aim for orthogonal API, and we can add shortcuts later.

Wait, did you mean bad idea instead of good idea here based on you saying we should keep it separate or orthogonal.

> I wonder if there's some precedent somewhere for such async builders?

hmm, yeah's let consider this for the release after the first one I'm making today or tomorrow. Would be interesting if we made the builder `await`able as that could replace the `transact()` calls

ChaoticTempest:
gonna pull this one in as it seems pretty robust in its current form, but we can still continue the conversation here

workspaces/src/network/account.rs
matklad: Using `AsRef` in the APIs is generally a bad idea, as it makes the function signature much harder to read, and also regresses type inference at the call-site. If the user have `xs: Vec<u8>`, they can just pass `&xs`. 

The only place where AsRef polymoprhism makes sense is for accepting a `&Path`. Note that it is needed there not to convert from owned to borrowed (PathBuf -> Path) but *between* borrowed types (&str -> &Path). And even than, it creates usability paper-cut, where passing in `PathBuf` works, but then you loose owenrship of it and can't, eg, print it in the error message.

Issue #18, opened 2021-10-19
Implement mainnet runtime


ChaoticTempest:
This will potentially be deferred out further, since this is not a main priority for now. https://github.com/near/workspaces-js does not have one either, but it would be nice down the road to have stuff working on testnet also be working on mainnet. Previous issue https://github.com/near/workspaces-rs/issues/3 already closed out the testnet portion.

If anyone needs this, feel free to bump this or make a comment.

PR #35, opened 2021-12-14
Add ref-finance example

ChaoticTempest:
- New ref-finance example to showcase a more in depth spooning example
- Adds mainnet network implementation to be used with the ref-finance example


ChaoticTempest:
This will also close out https://github.com/near/workspaces-rs/issues/18

workspaces/src/network/mainnet.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::network::Info;
+use crate::network::{
+    Account, CallExecution, CallExecutionDetails, NetworkClient, NetworkInfo,
+    TopLevelAccountCreator,
+};
+use crate::rpc::{client::Client, tool};
+use crate::types::{AccountId, InMemorySigner, SecretKey};
+use crate::Contract;
+
+const RPC_URL: &str = "https://rpc.mainnet.near.org";
```
TrevorJTClarke, line 18: Could/Should be env variable with fallback?

ChaoticTempest: Could be useful, but won't make it in this PR. Can you setup a new issue detailing it? Not sure what the best design to fit this in, but could be some form of configuration management down the line

TrevorJTClarke: yeah good call, ill do that

workspaces/src/network/mainnet.rs
```diff
@@ -0,0 +1,90 @@
...
+const HELPER_URL: &str = "https://helper.mainnet.near.org";
```
volovyk-s, line 19: Do we have a helper on the mainnet? Who is paying for that?

workspaces/src/network/mainnet.rs
```diff
@@ -0,0 +1,90 @@
...
+        sk: SecretKey,
+    ) -> anyhow::Result<CallExecution<Account>> {
+        tool::url_create_account(Url::parse(HELPER_URL)?, id.clone(), sk.public_key()).await?;
+        let signer = InMemorySigner::from_secret_key(id.clone(), sk);
+
+        Ok(CallExecution {
+            result: Account::new(id, signer),
+            details: CallExecutionDetails {
+                // We technically have not burnt any gas ourselves since someone else paid to
+                // create the account for us in testnet when we used the Helper contract.
```
volovyk-s, line 54: Probably you meant `mainnet`.

workspaces/src/network/mainnet.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl TopLevelAccountCreator for Mainnet {
+    async fn create_tla(
+        &self,
+        id: AccountId,
+        sk: SecretKey,
```
volovyk-s, line 45: Do we really need to pass SK here? Since we have this path `.near-credentials/mainnet/`, can we read it from the filesystem?

volovyk-s (commented):
As I understood, `mainnet` is used here only to clone contracts from it. For example in the `ref_finance` example. Why do we need helper and other functionality then? Are we really going to make transactions on `mainnet` in test env?

workspaces/src/network/mainnet.rs
ChaoticTempest: Aww man, didn't realize that mainnet doesn't have a helper contract to aid account creation. What's the flow look like for doing that for mainnet instead?

workspaces/src/network/mainnet.rs
ChaoticTempest: this function was just to be generic over how we could create top level accounts, because a user can grab credentials from just about anywhere. So if they wanted to read from `.near-credentials/mainnet/`, they have to do it manually themselves to be explicit or we offer a convenience function down the line

workspaces/src/network/mainnet.rs
TrevorJTClarke: you have to call the `near` contract directly to create a "sub account" on it.

Here's a snippet from nearapijs:
```
await creatorAccount.functionCall(
      'near',
      'create_account',
      {
        new_account_id: process.argv[3],
        new_public_key: process.argv[4]
      },
      '300000000000000',
      utils.format.parseNearAmount('0.1')
    )
    ```

ChaoticTempest:
> As I understood, mainnet is used here only to clone contracts from it. For example in the ref_finance example. Why do we need helper and other functionality then? Are we really going to make transactions on mainnet in test env?

The helper was copy-pasta'd and didn't realize mainnet didn't have one. `workspaces` should allow us to sign transactions for any network, but I can see how it be dangerous to do transactions on mainnet while doing tests, so probably best to have a feature flag for it. I'll make an issue to resolve that later since it could just snowball this PR

TrevorJTClarke:
> I'll make an issue to resolve that later since it could just snowball this PR

Our main use case is actually for automating deployments. Building a pipeline in testnet/guildnet allows full automation, then we want to rely on the same exact deployment flow in mainnet as well to make less human prone errors.

workspaces/src/network/mainnet.rs
ChaoticTempest: yeah, actually since mainnet requires an `Account` to pay the transaction of creating an `Account`, `create_tla` just won't be supported for mainnet. Instead, we'll have to have an existing `Account` on mainnet and do `Account::from_file()` then we can do `account.create_subaccount` to actually be able to create accounts

austinabell (approved):
LGTM, just added a few comments

PR #39, opened 2021-12-16
release: 0.1

ChaoticTempest:
After this PR, a release should be made for 0.1. So far, I've included the licenses, but I'm probably missing something essential. Let me know if there is something else I can include

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ This software is in very early alpha (use at your own risk). Only local sandboxe
 NOTE: since this is still alpha, we'll need to pull it down from github instead:
 ```
 [dependencies]
-workspaces = { git = "https://github.com/near/workspaces-rs", rev = "..." }
+workspaces = "0.1"
 ```
 Remember to grab the current revision since things are likely to change until published.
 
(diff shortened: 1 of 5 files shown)


volovyk-s:
@austinabell do we have an established process for the release of RS tools?
For JS libs we have this doc: https://www.notion.so/nearprotocol/App-layer-release-process-27ae5d35e1f44eb5b83184f100bf84f1
We are using `np` to automate the release. It checks the `git status`, runs tests, pushes git `TAG`s, and creates a release on GitHub.

austinabell:
> @austinabell do we have an established process for the release of RS tools? For JS libs we have this doc: [notion.so/nearprotocol/App-layer-release-process-27ae5d35e1f44eb5b83184f100bf84f1](https://www.notion.so/nearprotocol/App-layer-release-process-27ae5d35e1f44eb5b83184f100bf84f1) We are using `np` to automate the release. It checks the `git status`, runs tests, pushes git `TAG`s, and creates a release on GitHub.

We don't have a formally defined process, but I will make a document soon to try to setup some sort of process for consistency across repos. The thing about Rust releases is that the semver is quite specific and sensitive that handling the releases manually is the safest way to start until things stabilize imo

ChaoticTempest:
Probably should have the sandbox-utils crates in crates.io so we can pull it down from there instead of github. So https://github.com/near/sandbox/pull/20 should be merged and published first

austinabell (approved):

Issue #44, opened 2021-12-29
Cannot create several tests


NinoLipartiia4ire:
If you try running several separate tests with a sandboxed environment for the first time, at least one of them fails. 

See an example: https://github.com/NinoLipartiia4ire/workspaces-rs-example/commit/33a479d9f67c0fc6d2cdbc11510ccbc109f7a53e.
Both tests from this example should pass. But if you run them:
```bash
git clone https://github.com/NinoLipartiia4ire/workspaces-rs-example
cd workspaces-rs-example/tests-workspaces-rs && cargo test
```
Either one of them  fails:
```bash
test second_test ... ok
test first_test ... FAILED

failures:

---- first_test stdout ----
Starting up sandbox at localhost:24842
thread 'first_test' panicked at 'called `Result::unwrap()` on an `Err` value: Directory not empty (os error 66)', /Users/nino/.cargo/registry/src/github.com-1ecc6299db9ec823/workspaces-0.1.0/src/network/sandbox.rs:43:24
```
Or both tests fail:
```bash
test second_test ... FAILED
test first_test ... FAILED

failures:

---- second_test stdout ----
Starting up sandbox at localhost:22402
thread 'second_test' panicked at 'called `Result::unwrap()` on an `Err` value: /Users/nino/projects/issue-workspaces-rs/workspaces-rs-example/tests-workspaces-rs/target/debug/build/near-sandbox-utils-0a71fab0dd1ce1e6/out/near-sandbox-8b3f59a31cb9016b/near-sandbox is not executable', /Users/nino/.cargo/registry/src/github.com-1ecc6299db9ec823/workspaces-0.1.0/src/network/sandbox.rs:43:24
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- first_test stdout ----
Starting up sandbox at localhost:17497
thread 'first_test' panicked at 'called `Result::unwrap()` on an `Err` value: No such file or directory (os error 2)', /Users/nino/.cargo/registry/src/github.com-1ecc6299db9ec823/workspaces-0.1.0/src/network/sandbox.rs:43:24

```

Running `cargo test` for the second time and later doesn't return any error, both tests pass. 

To reproduce this bug again: `cargo clean && cargo test`.

ChaoticTempest:
I think I see the issue here. There was a race condition when installing the binary, so this should be fixed with this PR https://github.com/near/workspaces-rs/pull/46

mikedotexe:
[...]

PR #46, opened 2022-01-05
bugfix: Fixup race condition when installing near-sandbox

ChaoticTempest:
Move installing the near-sandbox binary into `build.rs` as to not contend with multiple tests trying to install it at once.

Fixes: https://github.com/near/workspaces-rs/issues/44

### DIFF
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -32,6 +32,9 @@ near-jsonrpc-primitives = "0.5"
 near-jsonrpc-client = { version = "0.1", features = ["sandbox"] }
 near-sandbox-utils = "0.1"
 
+[build-dependencies]
+near-sandbox-utils = "0.1"
+
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
 
--- a/workspaces/build.rs
+++ b/workspaces/build.rs
@@ -1,0 +1,3 @@
+fn main() {
+    near_sandbox_utils::install().expect("Could not install sandbox");
+}


austinabell (commented):
Won't this still fail if the env variable for the near bin location is unset on first compiling but then set when running? Is the race condition just that the second instance would try to use the binary before it is downloaded and initialized because it passed the check that it exists?

ChaoticTempest:
> Won't this still fail if the env variable for the near bin location is unset on first compiling but then set when running?

The env variable is only used for overriding the location of the binary in the case where the user wants to manually install `near-sandbox`, so won't matter if its unset then set when running.

>  Is the race condition just that the second instance would try to use the binary before it is downloaded and initialized because it passed the check that it exists?

it's slightly more interesting: from what I surmised, both instance would try to install at the same time, corrupting the binary. The `binary-install` already does the job of creating a temp folder and moves over the binary once fully downloaded and extracted, but it doesn't check whether or not the binary was correctly downloaded and only assumed that one instance would touch the temp folder

austinabell:
> it's slightly more interesting: from what I surmised, both instance would try to install at the same time, corrupting the binary. The `binary-install` already does the job of creating a temp folder and moves over the binary once fully downloaded and extracted, but it doesn't check whether or not the binary was correctly downloaded and only assumed that one instance would touch the temp folder

Then is there something that should be done around making these installations in different locations or having some lock on the file instead of just appending through the download? Is the issue with the move or the download into temp folder (is the temp folder the same?)

I just want to make sure we aren't just doing a band-aid fix that would uncover other issues later if we hit another race condition

ChaoticTempest:
> Is the issue with the move or the download into temp folder (is the temp folder the same?)

the temp folder is the same which is the issue

> I just want to make sure we aren't just doing a band-aid fix that would uncover other issues later if we hit another race condition

Good point. Just for good measure then, let me add in a lockfile to lock it from being attempting to install a second binary in the `near-sandbox` repo. Actually, is there a better way than just having a lockfile?

austinabell:
> Good point. Just for good measure then, let me add in a lockfile to lock it from being attempting to install a second binary in the `near-sandbox` repo. Actually, is there a better way than just having a lockfile?

What do you mean specifically by having a lockfile? Do you mean using something like `fs2` or equivalent, or something else?

ChaoticTempest:
yeah something like `fs2` or equivalent for locking a path or files within that path to not be modified

ChaoticTempest:
This should fix the issue, and I've opened up https://github.com/near/sandbox/pull/21 to address it on the sandbox installation side

PR #47, opened 2022-01-07
fix: Update Readme to updated and valid syntax

austinabell:
Some syntax wasn't valid, so updating so it's easier to follow

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ This software is in very early alpha (use at your own risk). Only local sandboxe
 - MacOS (x86) or Linux (x86) for sandbox tests. Testnet is available regardless
 
 ## Include it in our project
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,6 @@ This software is in very early alpha (use at your own risk). Only local sandboxe
 - MacOS (x86) or Linux (x86) for sandbox tests. Testnet is available regardless
 
 ## Include it in our project
-NOTE: since this is still alpha, we'll need to pull it down from github instead:
```
ChaoticTempest, line 11: thanks for noticing this one. I went to update the dependencies way back but somehow to get rid of the footnote

ChaoticTempest (approved):

PR #52, opened 2022-01-07
refactor!: don't require owned AccountId for RPC abstractions

austinabell:
There doesn't seem to me to be any requirement for these to be owned, given that they are just used to be serialized when sending a request. This is all very opinionated and I also have to think through the implications and future direction a lot more, but curious what you think @ChaoticTempest 

There were also some that could be changed with creating TLA (but this one seemed like owned might have a reason). Also, there are other parameters this could be done with, but trying to limit the scope.

### DIFF
--- a/examples/src/nft.rs
+++ b/examples/src/nft.rs
@@ -39,7 +39,7 @@ async fn main() -> anyhow::Result<()> {
     println!("nft_mint outcome: {:#?}", outcome);
 
     let result: serde_json::Value = worker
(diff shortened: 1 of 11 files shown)


ChaoticTempest (approved):
LGTM: it was originally requiring owned AccountId because the client mostly required it, but after a lot of changes it wasn't really needed to be owned anymore. Thanks for cleaning this up! Also the create TLA doesn't need to be owned either. All these types can be cheaply cloned either ways

Issue #53, opened 2022-01-11
Hide sandbox logs in tests by default


austinabell:
The running node logs should probably be hidden by default to not clutter the rust tests, especially since in the case of sandbox there is unlikely a log related to any TX. Would still be nice to be able to show/configure logs with `RUST_LOG` though.

Opening issue here for now, but I'm not sure if the logs can be suppressed or configured from running the process within the workspace test as it is.

PR #55, opened 2022-01-12
feat: Switch prints to `tracing`

### LINKED ISSUE #26: Switch prints to actual logging solution
Currently there are some `println` and `eprintln` calls, but we probably want these to be an actual logging solution so that devs can show/hide/filter or configure their logs.

The most general use case would be using the `log` crate I think, but if we are specifically tying our impl with `tokio`, we might want to use their `tracing` crate

itegulov:
Fixes #26 

I chose `tracing` since this is what `nearcore` is using and due to its extensive features such as layered spans and subscription mechanisms that can be useful for monitoring further down the line. As per conversation under the issue, it does not necessarily tie us to `tokio` ecosystem either.

Unfortunately, the sandbox logs are being produced by a dynamically downloaded binary file, so they do not respect the logging configuration (`RUST_LOG`, `RUST_LOG_STYLE` and others). See #53. This means that for now users will see logs being printed in an inconsistent manner like this:
```
2022-01-12T01:05:05.814561Z  INFO workspaces: Starting up sandbox at localhost:15562
2022-01-12T01:05:05.814601Z  INFO workspaces: Starting up sandbox at localhost:15965
Jan 12 12:05:05.817  INFO neard: Version: trunk, Build: 2c9375ee, Latest Protocol: 48
Jan 12 12:05:05.817  INFO neard: Version: trunk, Build: 2c9375ee, Latest Protocol: 48
Jan 12 12:05:05.817  INFO near: Generated node key, validator key, genesis file in /tmp/sandbox-15965
Jan 12 12:05:05.817  INFO near: Generated node key, validator key, genesis file in /tmp/sandbox-15562
2022-01-12T01:05:05.817696Z  INFO workspaces: Started sandbox: pid=12175
2022-01-12T01:05:05.817866Z  INFO workspaces: Started sandbox: pid=12176
```

### DIFF
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -23,6 +23,7 @@ serde = "1.0"
 serde_json = "1.0"
 tokio = { version = "1", features = ["full"] }
 tokio-retry = "0.3"
+tracing = "0.1.29"
 url = { version = "2.2.2", features = ["serde"] }
 
 near-account-id = "0.5"
@@ -40,3 +41,6 @@ libc = "0.2"
 
 [dev-dependencies]
 borsh = "0.9"
+env_logger = "0.9.0"
+test-log = { version = "0.2.8", default-features = false, features = ["trace"] }
+tracing-subscriber = {version = "0.3.5", default-features = false, features = ["env-filter", "fmt"]}
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -23,7 +23,7 @@ serde = "1.0"
 serde_json = "1.0"
 tokio = { version = "1", features = ["full"] }
 tokio-retry = "0.3"
-tracing = "0.1.29"
+tracing = "0.1"
 url = { version = "2.2.2", features = ["serde"] }
 
 near-account-id = "0.5"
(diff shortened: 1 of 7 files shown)


itegulov:
Also, I have not touched any of the examples, but it might be good idea to refactor at least one (or all?) of them to show that workspaces can be used with either `log` or `tracing`. @ChaoticTempest WDYT?

workspaces/Cargo.toml
```diff
@@ -23,6 +23,7 @@ serde = "1.0"
 serde_json = "1.0"
 tokio = { version = "1", features = ["full"] }
 tokio-retry = "0.3"
+tracing = "0.1.29"
```
austinabell, line 26:
```suggestion
tracing = "0.1"
```

Prob better for a lib like this to not have this as the minimum

austinabell (approved):

ChaoticTempest:
> it might be good idea to refactor at least one (or all?) of them to show that workspaces can be used with either

@itegulov Definitely a good idea, wanna try refactoring the spooning example? There's only a couple of `println` calls in there

itegulov:
@ChaoticTempest Done!

ChaoticTempest (approved):

Issue #49, opened 2022-01-07
ref-finance example fails to low storage deposit


austinabell:
error:

```
Error: Action #0: ExecutionError("Smart contract panicked: panicked at 'ERR_STORAGE_DEPOSIT need 3290000000000000000000, attatched 3000000000000000000000', ref-exchange/src/lib.rs:372:14")
```

Will investigate or add more details tomorrow, just wanted to open so I didn't forget

PR #57, opened 2022-01-19
Adds mainnet archival rpc and fixes ref-finance example

ChaoticTempest:
This will fix the issue with https://github.com/near/workspaces-rs/issues/49. This was because ref-finance contract on mainnet on updated and so did some deposit requirements get upped. With this PR, to remedy the issue, mainnet archival rpc was added so we can grab the contract from a specific block height instead of the latest block height

### DIFF
--- a/workspaces/src/rpc/patch.rs
+++ b/workspaces/src/rpc/patch.rs
@@ -1,14 +1,12 @@
 use near_crypto::KeyType;
 use near_jsonrpc_client::methods::sandbox_patch_state::RpcSandboxPatchStateRequest;
 use near_primitives::types::BlockId;
-use near_primitives::{
-    account::AccessKey, hash::CryptoHash, state_record::StateRecord, types::Balance,
-};
+use near_primitives::{account::AccessKey, state_record::StateRecord, types::Balance};
 
 use crate::network::DEV_ACCOUNT_SEED;
 use crate::rpc::client::Client;
-use crate::types::SecretKey;
-use crate::{AccountId, Contract, InMemorySigner};
+use crate::types::{BlockHeight, SecretKey};
+use crate::{AccountId, Contract, CryptoHash, InMemorySigner};
 
 pub struct ImportContractBuilder<'a, 'b> {
     account_id: AccountId,
@@ -41,8 +39,15 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
         }
     }
 
-    pub fn block_id(mut self, block_id: BlockId) -> Self {
-        self.block_id = Some(block_id);
+    pub fn block_height(mut self, block_height: BlockHeight) -> Self {
+        self.block_id = Some(BlockId::Height(block_height));
+        self
+    }
+
+    pub fn block_hash(mut self, block_hash: CryptoHash) -> Self {
+        self.block_id = Some(BlockId::Hash(near_primitives::hash::CryptoHash(
+            block_hash.0,
+        )));
         self
     }
 
@@ -57,13 +62,14 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
     }
 
     pub async fn transact(self) -> anyhow::Result<Contract> {
+        let account_id = self.account_id;
         let sk = SecretKey::from_seed(KeyType::ED25519, DEV_ACCOUNT_SEED);
         let pk = sk.public_key();
-        let signer = InMemorySigner::from_secret_key(self.account_id.clone(), sk);
+        let signer = InMemorySigner::from_secret_key(account_id.clone(), sk);
 
         let mut account_view = self
             .from_network
-            .view_account(self.account_id.clone(), self.block_id.clone())
+            .view_account(account_id.clone(), self.block_id.clone())
             .await?;
         if let Some(initial_balance) = self.initial_balance {
             account_view.amount = initial_balance;
@@ -71,23 +77,23 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
 
         let mut records = vec![
             StateRecord::Account {
-                account_id: self.account_id.clone(),
+                account_id: account_id.clone(),
                 account: account_view.clone().into(),
             },
             StateRecord::AccessKey {
-                account_id: self.account_id.clone(),
+                account_id: account_id.clone(),
                 public_key: pk.clone().into(),
                 access_key: AccessKey::full_access(),
             },
         ];
 
-        if account_view.code_hash != CryptoHash::default() {
+        if account_view.code_hash != near_primitives::hash::CryptoHash::default() {
             let code_view = self
                 .from_network
-                .view_code(self.account_id.clone(), self.block_id.clone())
+                .view_code(account_id.clone(), self.block_id.clone())
                 .await?;
             records.push(StateRecord::Contract {
-                account_id: self.account_id.clone(),
+                account_id: account_id.clone(),
                 code: code_view.code,
             });
         }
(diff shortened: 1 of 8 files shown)


volovyk-s (approved):

workspaces/src/types.rs
```diff
@@ -5,6 +5,7 @@ use std::path::Path;
 
 pub use near_account_id::AccountId;
 pub(crate) use near_crypto::{KeyType, Signer};
+pub use near_primitives::types::BlockId;
```
austinabell, line 8: Perhaps we should have a local type to not be affected by the updates to near_primitives as well as making ours more ergonomic/better docs and just swap the type internally?

workspaces/src/rpc/patch.rs
```diff
@@ -86,7 +95,7 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
         if self.import_data {
             records.extend(
                 self.from_network
-                    .view_state_raw(self.account_id.clone(), None, None)
+                    .view_state_raw(self.account_id.clone(), None, self.block_id.clone())
```
austinabell, line 98: very optional, but can avoid the clone here and simplify the code very slightly:

```
diff --git a/workspaces/src/rpc/patch.rs b/workspaces/src/rpc/patch.rs
index a459fec..96cc5c2 100644
--- a/workspaces/src/rpc/patch.rs
+++ b/workspaces/src/rpc/patch.rs
@@ -57,13 +57,14 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
     }
 
     pub async fn transact(self) -> anyhow::Result<Contract> {
+        let account_id = self.account_id;
         let sk = SecretKey::from_seed(KeyType::ED25519, DEV_ACCOUNT_SEED);
         let pk = sk.public_key();
-        let signer = InMemorySigner::from_secret_key(self.account_id.clone(), sk);
+        let signer = InMemorySigner::from_secret_key(account_id.clone(), sk);
 
         let mut account_view = self
             .from_network
-            .view_account(self.account_id.clone(), self.block_id.clone())
+            .view_account(account_id.clone(), self.block_id.clone())
             .await?;
         if let Some(initial_balance) = self.initial_balance {
             account_view.amount = initial_balance;
@@ -71,11 +72,11 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
 
         let mut records = vec![
             StateRecord::Account {
-                account_id: self.account_id.clone(),
                 account: account_view.clone().into(),
             },
             StateRecord::AccessKey {
-                account_id: self.account_id.clone(),
                 public_key: pk.clone().into(),
                 access_key: AccessKey::full_access(),
             },
@@ -84,10 +85,10 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
         if account_view.code_hash != CryptoHash::default() {
             let code_view = self
                 .from_network
-                .view_code(self.account_id.clone(), self.block_id.clone())
+                .view_code(account_id.clone(), self.block_id.clone())
                 .await?;
             records.push(StateRecord::Contract {
-                account_id: self.account_id.clone(),
                 code: code_view.code,
             });
         }
@@ -95,11 +96,11 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
         if self.import_data {
             records.extend(
                 self.from_network
-                    .view_state_raw(self.account_id.clone(), None, self.block_id.clone())
+                    .view_state_raw(account_id.clone(), None, self.block_id)
                     .await?
                     .into_iter()
                     .map(|(key, value)| StateRecord::Data {
-                        account_id: self.account_id.clone(),
+                        account_id: account_id.clone(),
                         data_key: key,
                         value,
                     }),
@@ -120,6 +121,6 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
             .await
             .map_err(|err| anyhow::anyhow!("Failed to patch state: {:?}", err))?;
 
-        Ok(Contract::new(self.account_id, signer))
+        Ok(Contract::new(account_id, signer))
     }
 }

```

workspaces/src/rpc/patch.rs
```diff
@@ -34,15 +37,21 @@ impl<'a, 'b> ImportContractBuilder<'a, 'b> {
...
             import_data: false,
             initial_balance: None,
+            block_id: None,
         }
     }
 
+    pub fn block_id(mut self, block_id: BlockId) -> Self {
+        self.block_id = Some(block_id);
+        self
+    }
```
austinabell, line 47: To abstract away this type, perhaps we can have `block_height` and `block_hash` be separate to avoid having to export the `BlockId` type as before? Unless we already need to expose the block height for the query outside of the builder pattern?

ChaoticTempest: interesting, but we'll need to export `CryptoHash` otherwise though. How about we just abstract away this type anyways so people don't have to import `BlockId` and only need to import `CryptoHash` if they need it for referencing specific block hashes, so that block height version doesn't need the import at all

workspaces/src/rpc/patch.rs
ChaoticTempest: good eye, I'll try that out. Thanks!

workspaces/src/types.rs
ChaoticTempest: Let's go with the version you mentioned later about abstracting it

workspaces/src/rpc/patch.rs
austinabell: CryptoHash is just an alias, so less of an issue to import, but we could just have our own local type for this with no difference, as an option

workspaces/src/types.rs
```diff
@@ -55,3 +59,37 @@ impl InMemorySigner {
         &self.0
     }
 }
+
+// type taken from near_primitives::hash::CryptoHash.
+/// CryptoHash is type for storing the hash of a specific block.
+pub struct CryptoHash(pub [u8; 32]);
```
austinabell, line 65: why use newtype for this over just an alias of `[u8; 32]` for simplicity?

ChaoticTempest: I just took whatever was in `near_primitives`. It's a newtype there, so stuck with that. I don't know if changing it to an alias of `[u8; 32]` would make it harder to transition later when `near_primitives` is `1.0`

austinabell: why would it be hard to transition? Just wrapping the bytes passed in with `CryptoHash` internally? Am I missing something with what you mean here?

ChaoticTempest: I'm just thinking whether or not if we end up just exposing `near_primitives::hash::CryptoHash` later and its still just a newtype which would make it a breaking change to transition over to. But maybe not a huge concern if we force everyone to just rely on our `CryptoHash` from workspaces

austinabell: yeah, the primitives type prob shouldn't be tied to our API. I'm not sure if I'm misunderstanding what you're saying here though. Why would we ever need to expose the primitives type to be a part of our API?

ChaoticTempest: I was just thinking of this case, because there was a conversation we had about potentially exposing the `near_primitives`/`near_account_id` types once they released 1.0. But regardless of that, I think the newtype here would be better suited since it might be harder for users to create `CryptoHash`es if it was just an alias of `[u8; 32]`. The `FromStr` impl should allow us to create it easily

ChaoticTempest:
I'll just merge this one in since this one has a fix for the other PRs stalling. If there's anymore concerns, we can just address them in a later ticket/PR

PR #60, opened 2022-01-25
fix: Make `access_key` do an optimistic RPC query

itegulov:
This PR aligns `access_key` to behave like all other view queries and make an optimistic (finality-wise) RPC query. This is especially important since `access_key` is also used to make function call transactions, so if you make many calls in a quick succession this can lead to duplicate access key nonces being returned by `access_key`.


ChaoticTempest (approved):
Thanks again for catching it

PR #70, opened 2022-02-04
refactor: `CallResultDetails` now expose logs, transaction and receipt outcomes

ChaoticTempest:
I went with exposing `near-primitives::views::ExecutionStatusView` as-is just like how we're exposing `near-account-id::AccountId` as-is. There's a lot of implementation that would have to be moved over such as `TxExecutionError` type. This might be hard to maintain in the future as the near-primitives crate gets updated. So I think it's best if we just expose it as-is for now since this is just a view type where the users can grab the data. @austinabell WDYT?

Something to consider: trying to rethink the `CallExecution` type and see whether it's necessary. One of the cases where we're using it with `anyhow::Result<CallExecution<Account>>` leads to nested results which isn't ideal for usage. Maybe it's better if we go with something like `Result<Account, CallExecutionError>` where we convert the `CallExecutionDetails` into an error type instead where we can also store all the execution info


workspaces/src/network/result.rs
```diff
@@ -81,6 +86,60 @@ impl CallExecutionDetails {
...
+    /// from the transaction and all the receipts it generated.
+    pub fn outcomes(&self) -> Vec<&ExecutionOutcome> {
+        let mut outcomes = vec![&self.transaction];
+        outcomes.extend(&self.receipts);
+        outcomes
+    }
+
+    /// Grab all outcomes after the execution of the transaction. This includes outcomes
+    /// only from receipts generated by this transaction.
+    pub fn receipt_outcomes(&self) -> &Vec<ExecutionOutcome> {
```
austinabell, line 105:
```suggestion
    pub fn receipt_outcomes(&self) -> &[ExecutionOutcome] {
```

workspaces/src/network/result.rs
```diff
@@ -81,6 +86,60 @@ impl CallExecutionDetails {
...
+    }
+
+    /// Grab all outcomes that did not succeed the execution of this transaction. This
+    /// will also include the failures from receipts as well.
+    pub fn failures(&self) -> Vec<&ExecutionOutcome> {
+        let mut failures = Vec::new();
+        if matches!(self.transaction.status, ExecutionStatusView::Failure(_)) {
+            failures.push(&self.transaction);
+        }
+        failures.extend(self.receipt_failures());
```
austinabell, line 116: I know optimizing the allocations for this isn't really that useful since it's scripting, but perhaps could use receipt failures length for the vec capacity, or even for the case of the primary transaction not succeeding to just return this to avoid copying and re-allocating?

Honestly not sure if this can be optimized by the compiler

workspaces/src/network/result.rs
```diff
@@ -81,6 +86,60 @@ impl CallExecutionDetails {
...
+    /// Just like `failures`, grab only failed receipt outcomes.
+    pub fn receipt_failures(&self) -> Vec<&ExecutionOutcome> {
+        self.receipts
+            .iter()
+            .filter(|receipt| matches!(receipt.status, ExecutionStatusView::Failure(_)))
+            .collect()
+    }
+
+    /// Checks whether the transaction was successful. Returns true if
+    /// the transaction has a status of FinalExecutionStatus::Success.
```
austinabell, line 129:
```suggestion
    /// the transaction has a status of [FinalExecutionStatus::Success].
```
Prob worth having intra doc links for these (I didn't check if this is resolved, so you might need to fix if this doesn't)

workspaces/src/network/result.rs
```diff
@@ -81,6 +91,60 @@ impl CallExecutionDetails {
...
+    pub fn is_failure(&self) -> bool {
+        matches!(self.status, FinalExecutionStatus::Failure(_))
+    }
+
+    /// Grab all logs from both the transaction and receipt outcomes.
+    pub fn logs(&self) -> Vec<String> {
+        self.outcomes()
+            .iter()
+            .flat_map(|outcome| &outcome.logs)
+            .map(Clone::clone)
```
austinabell, line 145:
```suggestion
            .cloned()
```

workspaces/src/network/result.rs
```diff
@@ -81,6 +91,60 @@ impl CallExecutionDetails {
...
+    }
+
+    /// Checks whether the transaction has failed. Returns true if
+    /// the transaction has a status of FinalExecutionStatus::Failure.
+    pub fn is_failure(&self) -> bool {
+        matches!(self.status, FinalExecutionStatus::Failure(_))
+    }
+
+    /// Grab all logs from both the transaction and receipt outcomes.
+    pub fn logs(&self) -> Vec<String> {
```
austinabell, line 141: Are we sure we want cloned logs? Can we not return references to the logs as `Vec<&str>`?

workspaces/src/network/result.rs
```diff
@@ -93,6 +157,12 @@ impl From<FinalExecutionOutcomeView> for CallExecutionDetails {
                     .iter()
                     .map(|t| t.outcome.gas_burnt)
                     .sum::<u64>(),
+            transaction: transaction_result.transaction_outcome.into(),
+            receipts: transaction_result
+                .receipts_outcome
+                .into_iter()
+                .map(Into::into)
```
austinabell, line 164:
```suggestion
                .map(ExecutionOutcome::from)
```

Optional: Might be more readable if these conversion types are specified here

workspaces/src/types.rs
```diff
@@ -2,11 +2,30 @@
...
+    pub use near_primitives::errors::{
+        ActionError, ActionErrorKind, InvalidAccessKeyError, InvalidTxError, TxExecutionError,
+    };
+}
+
+/// Status module exposing types where we can view the status of a transaction's
+/// outcome.
+pub mod status {
+    pub use near_primitives::views::{ExecutionStatusView, FinalExecutionStatus};
+}
```
austinabell, line 20: I'm skeptical it's the best long-term plan to expose all of these types to the API, seems like it would just be a maintenance pain to do so. I don't have a great grasp on why this is needed just yet

austinabell:
> I went with exposing `near-primitives::views::ExecutionStatusView` as-is just like how we're exposing `near-account-id::AccountId` as-is. There's a lot of implementation that would have to be moved over such as `TxExecutionError` type. This might be hard to maintain in the future as the near-primitives crate gets updated. So I think it's best if we just expose it as-is for now since this is just a view type where the users can grab the data. @austinabell WDYT?
> 
> Something to consider: trying to rethink the `CallExecution` type and see whether it's necessary. One of the cases where we're using it with `anyhow::Result<CallExecution<Account>>` leads to nested results which isn't ideal for usage. Maybe it's better if we go with something like `Result<Account, CallExecutionError>` where we convert the `CallExecutionDetails` into an error type instead where we can also store all the execution info

I definitely need to look at the API as a whole tomorrow (or another day) to give an opinion on this.

workspaces/src/network/result.rs
ChaoticTempest: ahh was just being a dumb dumb on this one. Was thinking `&String` at first too haha

workspaces/src/types.rs
ChaoticTempest: For context, this would be needed if a user wants to grab data from the status of an outcome and match it to a specific error type to do whatever they wish with it. We already offer a `failures()` API, but that might not suit all cases.

The alternative could just be forcing users to use `near_primitives::errors::*` instead, but think we had the discussion of that not being so great either. The error types are pretty big, but I'd imagine exposing them to not hurt us so much since eventually `near_primitives` 1.0 should be released. @miraclx do you have any opinions on this?

workspaces/src/network/result.rs
willemneal: Shouldn't this return an Option or Result type? That would avoid the allocation in the negative case.

ChaoticTempest: I don't think this function needs to be that optimized. Would probably be better to keep it consistent w/ the other APIs here and return something that is iterable even if we make the allocation. The negative case still works with it just being an empty vec

workspaces/src/types.rs
miraclx: > This might be hard to maintain in the future as the near-primitives crate gets updated. So I think it's best if we just expose it as-is for now since this is just a view type where the users can grab the data.

This sounds good to me. Let's not expose the users directly to a crate, whose API is certain to change a lot very soon. Instead, by providing those types from within workspaces, we abstract over them. At least until the new crate is ready. At that point, if we decide it's the better way forward, we can drop them from workspaces and users can migrate to using that directly.

workspaces/src/network/result.rs
```diff
@@ -133,3 +205,84 @@ impl From<CallResult> for ViewResultDetails {
...
+    /// to an [`anyhow::Error`] object which can be downcasted later.
+    pub fn into_result(self) -> anyhow::Result<ValueOrReceiptId> {
+        match self.status {
+            ExecutionStatusView::SuccessValue(value) => Ok(ValueOrReceiptId::Value(value)),
+            ExecutionStatusView::SuccessReceiptId(hash) => {
+                Ok(ValueOrReceiptId::ReceiptId(CryptoHash(hash.0)))
+            }
+            ExecutionStatusView::Failure(err) => {
+                Err(anyhow::anyhow!("Execution failed: {:?}", err))
+            }
```
ChaoticTempest, line 257: Will need to directly change this formatted anyhow error in the future when https://github.com/near/nearcore/pull/6367 makes it into a near_primitives release

workspaces/src/types.rs
ChaoticTempest: @austinabell The errors are now abstracted behind anyhow ones and are no longer exposed. Let me know what you think

workspaces/src/types.rs
```diff
@@ -2,11 +2,16 @@
 /// and internal libraries like near-jsonrpc-client requires specific versions
 /// of these types which shouldn't be exposed either.
 use std::convert::TryFrom;
+use std::fmt;
 use std::path::Path;
 
 pub use near_account_id::AccountId;
 pub(crate) use near_crypto::{KeyType, Signer};
-use near_primitives::serialize::from_base;
+pub use near_primitives::views::ExecutionStatusView;
```
austinabell, line 10: Seems not ideal to expose this type from primitives. There are implementations and details of this that might be tricky to maintain, and the types within this enum do not match anything from this lib

ChaoticTempest: Yup, thanks for catching that. Thought this was cleaned up when I removed all the exports but guess not

Issue #67, opened 2022-02-02
Add support for batched transactions


itegulov:
NEAR runtime allows one to submit multiple actions batched into one transaction (see https://nomicon.io/RuntimeSpec/Transactions#batched-transaction). Would be nice to have access to this functionality from workspaces.

PR #72, opened 2022-02-08
feat: batch transactions

ChaoticTempest:
addresses: https://github.com/near/workspaces-rs/issues/67.

This PR adds a new `Transaction` type that allows us to add in actions into it, to form a batched transaction. A lot of stuff such as `client.rs` can use it directly instead of making individual calls, so that it's consistent throughout, but will save that for later in another PR.

One thing to note is that `batch_tx.call()` takes in a `CallArgs` type just so we can be consistent with the `account.call().args().deposit()...` API, and not need to specify optional parameters such as gas/deposit. This is how it would look like when adding calls into a batch transaction:
```rust
account.batch_tx(&worker, contract_id)
    .call(CallArgs::new("function")
        .args(...)
        .deposit(0)
    )
    ...
    .transact()
    .await?;
```
Not sure if this API is the best, or if there's a better naming than `batch_tx`. Let me know if there is.

Also considered the following API for `.call` but probably is too funky. Also would require a trait to be exposed in the prelude:
```rust
.call("function_name".into_call_args()
    .args(...)
    .deposit(...)
)
```


workspaces/tests/batch_tx.rs
```diff
@@ -0,0 +1,41 @@
...
+        .call(
+            CallArgs::new("set_status")
+                .args_json(json!({
+                    "message": "hello_world",
+                }))?
+                .deposit(0),
+        )
+        .call(CallArgs::new("set_status").args_json(json!({
+            "message": "world_hello",
+        }))?)
```
itegulov, line 26: The stylistic discrepancy between these two calls looks pretty weird, but I assume this is because of rustfmt...

In general I am not a big fan of how this "reads", but at the same time I can't come up with anything better right now.

workspaces/src/network/account.rs
```diff
@@ -174,59 +187,60 @@ impl Contract {
     ) -> anyhow::Result<CallExecutionDetails> {
         self.account.delete_account(worker, beneficiary_id).await
     }
+
+    /// Start a batch transaction, using the current contract as the signer and
+    /// making calls into this contract. Returns a [`Transaction`] object that
+    /// we can use to add Actions to the batched transaction. Call `transact`
+    /// to send the batched transaction to the network.
+    pub fn batch_tx<'a, T: Network>(&self, worker: &'a Worker<T>) -> Transaction<'a> {
```
itegulov, line 195: I think I would prefer this function to be called `call_batch` to stay consistent with other call-"initiating" functions. We will have to call `.transact()` in the end anyway, so the transactional nature is implied here.

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,159 @@
...
+        })
+    }
+}
+
+pub struct Transaction<'a> {
+    client: &'a Client,
+    signer: InMemorySigner,
+    receiver_id: AccountId,
+    actions: Vec<Action>,
+}
```
itegulov, line 74: Shouldn't this be called something like `TransactionBuilder` to stay consistent with `CallBuilder`? Also, expanding on the topic above, `BatchCallBuilder` might be more suitable?

itegulov (commented):
LGTM in general, just some naming nitpicking for now. I will make an attempt to use it in some of the examples later in the week and let you know my feedback.

workspaces/src/network/transaction.rs
ChaoticTempest: yeah you're right that it should be consistent, but `CallBuilder` should be renamed to `CallTransaction` instead, since every other API such as near-api-js and near-primitives defines this `Transaction` type where you add actions. So think its best to stay consistent with those and just rename `CallBuilder` instead

workspaces/src/network/account.rs
ChaoticTempest: actually this might not be a great idea for naming either calls are more associated to making calls into contract functions, which doesn't suit all cases this batch transaction is meant for such as staking or account creation:
```rust
account.call_batch(new_account_id)
    .create_account()
    .transfer(...)
    .stake(...)
```

itegulov: Good point, yeah. In that case, I would go for either your original option or for a simple `batch`. Regardless, I do not really have a strong opinion here.

ChaoticTempest: Renamed to `batch` instead since the `tx` is kinda redundant at the end of the day

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    /// Process the trannsaction, and return the result of the execution.
+    pub async fn transact(self) -> anyhow::Result<CallExecutionDetails> {
+        self.transact_raw().await.map(Into::into)
+    }
+}
+
+/// Similiar to a [`Transaction`], but more specific to making a call into a contract.
+/// Note, only one call can be made per `CallTransaction`.
+pub struct CallTransaction<'a, T> {
```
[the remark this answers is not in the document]
ChaoticTempest: I don't think it'll be the most intuitive to use for a single `call` function since they're not really the same APIs. Also would need to import `Function` every time now.

```rust
contract.call(
    worker,
    Function::new(name)
       .args_json(...)
)
.transact();
```

Which looks kinda funky over just:
```rust
contract.call(worker, name)
     .args_json(...)
     .transact()
```

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,323 @@
...
+use crate::network::{CallExecution, CallExecutionDetails, Network, ViewResultDetails};
+use crate::rpc::client::{
+    send_batch_tx_and_retry, Client, DEFAULT_CALL_DEPOSIT, DEFAULT_CALL_FN_GAS,
+};
+use crate::types::{AccessKey, AccountId, Balance, Gas, InMemorySigner, PublicKey, SecretKey};
+use crate::worker::Worker;
+use crate::Account;
+
+#[derive(Debug, Clone)]
+pub struct CallArgs {
```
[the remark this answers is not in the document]
ChaoticTempest: that definitely reads better for sure

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,325 @@
...
+    pub fn args(mut self, args: Vec<u8>) -> Self {
+        self.call_args = self.call_args.args(args);
+        self
+    }
+
+    /// Similiar to `args`, specify an argument that is JSON serializable and can be
+    /// accepted by the equivalent contract. Recommend to use something like
+    /// `serde_json::json!` macro to easily serialize the arguments.
+    pub fn args_json<U: serde::Serialize>(mut self, args: U) -> anyhow::Result<Self> {
+        self.call_args = self.call_args.args_json(args)?;
```
miraclx, line 203:
```suggestion
        self.function = self.function.args_json(args)?;
```

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,325 @@
...
+        Self::FunctionCall(FunctionCallAction {
+            method_name: args.function,
+            args: args.args,
+            deposit: args.deposit,
+            gas: args.gas,
+        })
+    }
+}
+
+pub struct Transaction<'a> {
```
miraclx, line 75: wdyt about naming this something like `BatchTransaction`. Esp since we have other types of `Transaction`-s: `CallTransaction`, `CreateAccountTransaction`.

miraclx (approved):
lgtm.

workspaces/src/network/transaction.rs
ChaoticTempest: I didn't go with that because Transactions are BatchTransactions. Like there's just a bunch of actions in a single transaction, plus other places such as near-api-js and nearcore call it purely just `Transaction`, so thought it be best to stay consistent that way

workspaces/src/types.rs
```diff
@@ -93,3 +97,101 @@ impl TryFrom<Vec<u8>> for CryptoHash {
...
+    pub fn full_access() -> Self {
+        Self {
+            nonce: 0,
+            permission: AccessKeyPermission::FullAccess,
+        }
+    }
+
+    pub fn function_call_access(
+        receiver_id: &AccountId,
+        method_names: Vec<String>,
```
austinabell, line 127: should be a slice and only require `&str` as type?

austinabell (commented):
just a quick scan for now. Will dig more into it later since it seems like there can be some duplication removed, but I need to play around with it to see what is possible

workspaces/src/network/transaction.rs
miraclx: Agreed. It's just because we have different types of `Transactions`, I thought it'll make sense to distinguish. I will mark this as resolved.

workspaces/src/types.rs
```diff
@@ -124,14 +124,14 @@ impl AccessKey {
 
     pub fn function_call_access(
         receiver_id: &AccountId,
-        method_names: Vec<String>,
+        method_names: Vec<&str>,
```
austinabell, line 127:
```suggestion
        method_names: &[&str],
```
?

ChaoticTempest: yeah that works better!

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,323 @@
...
+};
+use crate::types::{AccessKey, AccountId, Balance, Gas, InMemorySigner, PublicKey, SecretKey};
+use crate::worker::Worker;
+use crate::Account;
+
+/// A set of arguments we can provide to a transaction, containing
+/// the function name, arguments, the amount of gas to use and deposit.
+#[derive(Debug, Clone)]
+pub struct Function {
+    name: String,
```
austinabell, line 22: probably worth changing this to be a ref and have a lifetime now. This will be changed in the future to have one so no need to clone early to avoid

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,323 @@
...
+        Self::FunctionCall(FunctionCallAction {
+            method_name: function.name,
+            args: function.args,
+            deposit: function.deposit,
+            gas: function.gas,
+        })
+    }
+}
+
+pub struct Transaction<'a> {
```
austinabell, line 75: Can you document this? I think it isn't intuitive the ambiguity mentioned by @miraclx but also that this is a builder pattern that is generated through something else

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    /// Create a new account with the account id being `receiver_id`.
+    pub fn create_account(mut self) -> Self {
+        self.actions.push(CreateAccountAction {}.into());
+        self
+    }
+
+    /// Deletes the `receiver_id`'s account. The beneficiary specified by
+    /// `beneficiary_id` will receive the funds of the account deleted.
+    pub fn delete_account(mut self, beneficiary_id: AccountId) -> Self {
```
austinabell, line 119: nothing potentially to change now, but these parameters might in the future be a reference to the `AccountId`, or is that not possible here for some reason? Asking to potentially avoid a breaking change down the road.

workspaces/src/network/transaction.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    /// Specify the amount of tokens to be deposited where `deposit` is the amount of
+    /// tokens in yocto near.
+    pub fn deposit(mut self, deposit: u128) -> Self {
+        self.function = self.function.deposit(deposit);
+        self
+    }
+
+    /// Specify the amount of gas to be used where `gas` is the amount of gas in yocto near.
+    pub fn gas(mut self, gas: u64) -> Self {
```
austinabell, line 224: Can you use the Gas and Balance aliases within types for these?

workspaces/src/types.rs
```diff
@@ -11,6 +11,10 @@ use near_primitives::logging::pretty_hash;
 use near_primitives::serialize::{from_base, to_base};
 use serde::{Deserialize, Serialize};
 
+/// Nonce for transactions.
+pub type Nonce = u64;
+
+/// Gas is a type for storing amount of gas.
```
austinabell, line 17:
```suggestion
/// Gas units that are used to execute operations for a transaction.
```

I know this comment came from primitives, but it's not great. Feel free to re-word, but this is my stab at a better comment

workspaces/tests/batch_tx.rs
```diff
@@ -0,0 +1,41 @@
...
+        .call(
+            Function::new("set_status")
+                .args_json(json!({
+                    "message": "hello_world",
+                }))?
+                .deposit(0),
+        )
+        .call(Function::new("set_status").args_json(json!({
+            "message": "world_hello",
+        }))?)
```
austinabell, line 26:
```suggestion
        .call("set_status")
        .args_json(json!({
            "message": "hello_world",
         }))?
         .deposit(0)
        .call("set_status")
        .args_json(json!({
            "message": "world_hello",
        }))?
```

This type of pattern seems more intuitive for me, but I can see that it could create a more complex builder type (needing to have an in progress function with batch and not) but also doesn't seem that much more complex than having the `Function` builder type having to be built manually.

Is this something you've explored and decided against? Would it be helpful if I made a commit with this change to compare, or do you understand what my suggestion is?

workspaces/tests/batch_tx.rs
```diff
@@ -0,0 +1,41 @@
...
+async fn test_batch_tx() -> anyhow::Result<()> {
+    let worker = workspaces::sandbox();
+    let contract = worker
+        .dev_deploy(include_bytes!("../../examples/res/status_message.wasm"))
+        .await?;
+
+    // Batch transaction with two `call`s into `set_status`. The second one
+    // should override the first one.
+    contract
+        .batch(&worker)
```
austinabell, line 16: perhaps `batch_transaction` might be more clear for a function name?

workspaces/src/network/transaction.rs
ChaoticTempest: good catch, a lot of this was generated code from github copilot so didn't catch this at first glance or the type aliases you mentioned in another comment

workspaces/tests/batch_tx.rs
ChaoticTempest: Yeah I explored this one already, and there's one part to this that didn't make sense: the above builder syntax would violate type state pattern where we're allowing "non-call" functions to be used with `call` specifying arguments like so:
```rust
account.batch(&worker)
     .create_account()
     .deposit(...)
     .transact()
```
which could be confusing since users might think `deposit` is setting the deposit/transfer-amount into the `create_account` call. So went against this pattern for that reason

workspaces/tests/batch_tx.rs
ChaoticTempest: I originally had `batch_tx` but thought it might've been redundant since we call `.transact()` at the end anyways which should signify that it's a transaction. And also was thinking that would be better for consistency since we don't have `call_transaction` either

workspaces/src/network/transaction.rs
ChaoticTempest: hmm, I'm not sure I see why it would get changed in the future? Is there something that's getting changed somewhere that I'm missing? `Function` should be donating its internals later on anyways, so there shouldn't be a wasted clone/allocation here

workspaces/tests/batch_tx.rs
austinabell: that's fair!

workspaces/tests/batch_tx.rs
ChaoticTempest: mmm, I see. So you mean something like this: 
```rust
account
   .batch(&worker)
   .create_account()
   .transfer("1 N")
   .deploy("path/wasm")
   .call("x")
   .args_json((1, 2, 3))
   .deposit(1000)
   .stake("0.5 N")
   .call("y")
   .args_json((4, 5, 6))
   .transact()
```
That could work, but now it might feel like we're breaking the flow with mixing the call argument functions in with the rest of the transaction functions. I like it overall, but still feel a little hesitant with how it flows. What do you think? Or which would you prefer, the above or the following:
```rust
account
   .batch(&worker)
   .create_account()
   .transfer("1 N")
   .deploy("path/wasm")
   .call(
       Function::new("x")
          .args_json((1, 2, 3))
          .deposit(1000)
    )
   .stake("0.5 N")
   .call(
       Function::new("x")
          .args_json((4, 5, 6))
   )
   .transact()
```

austinabell: To be clear, I'm totally fine with what there was, just wanted to offer an alternative suggestion since it seemed a bit unergonomic to use for a script like this

workspaces/src/network/transaction.rs
ChaoticTempest: this is all just internal though, and I meant like we end up converting a `Function` over to a `FunctionCallAction` later on which would take an owned string in its current form. But potentially could be changed in the future too. Not too big of a deal but I'll change it over to the ref version

workspaces/tests/batch_tx.rs
ChaoticTempest: ended up just keeping the former, but we can go back to this if we get enough feedback that doesn't like pulling in `Function` to use with `call`s

austinabell (approved):

PR #73, opened 2022-02-09
feat: add fast-forwarding

ChaoticTempest:
Adds the `worker.fast_forward(x)` specific API for simulating producing `x` number of blocks on the sandbox node. Also adds a simple example for fast forwarding and grabbing time specific data related to how many blocks have been produced before and after a `fast_forward` request.

Mainly opening this one up for comments so we can review this one sooner than later, and will point to the specific nearcore version once https://github.com/near/nearcore/pull/6211 merges


workspaces/Cargo.toml
```diff
@@ -26,15 +26,15 @@ tokio-retry = "0.3"
...
-near-crypto = "0.5"
-near-primitives = "0.5"
-near-jsonrpc-primitives = "0.5"
-near-jsonrpc-client = { version = "0.1", features = ["sandbox"] }
-near-sandbox-utils = "0.1.1"
+near-account-id = "0.12"
+near-crypto = "0.12"
+near-primitives = "0.12"
+near-jsonrpc-primitives = "0.12"
+near-jsonrpc-client = { git = "https://github.com/near/near-jsonrpc-client-rs", features = ["sandbox"] }
```
miraclx, line 33: `near-jsonrpc-client` `v0.3.0` has been released. https://github.com/near/near-jsonrpc-client-rs/releases/tag/v0.3.0


```suggestion
near-jsonrpc-client = { version = "0.3", features = ["sandbox"] }
```

miraclx (changes requested):

workspaces/src/network/mod.rs
```diff
@@ -151,6 +154,18 @@ where
     ) -> ImportContractTransaction<'a, 'b> {
         ImportContractTransaction::new(id.to_owned(), worker.client(), self.client())
     }
+
+    async fn fast_forward(&self, delta_height: u64) -> anyhow::Result<()> {
+        // NOTE: RpcSandboxFastForwardResponse is an empty struct with no fields, so don't do anything with it:
+        let _forward_resp = self
```
austinabell, line 160:
```suggestion
        self
```

If you're not checking the response serialization is correct, probably no need to bind this to a variable

workspaces/src/worker/impls.rs
```diff
@@ -77,6 +77,10 @@ where
     ) -> ImportContractTransaction<'a, 'b> {
         self.workspace.import_contract(id, worker)
     }
+
+    async fn fast_forward(&self, delta_height: u64) -> anyhow::Result<()> {
```
austinabell, line 81: noting that adding this function to `StatePatcher` blurs that naming meaning. These things seem sandbox specific so not sure why the trait is needed yet

austinabell (approved):

workspaces/src/network/mod.rs
ChaoticTempest: yeah, was binding to a variable to point out to readers that the resp object is empty, but it's not a big deal, so I'll just get rid of it

austinabell (approved):
Didn't notice that you moved the logic similar to #102 for removing StatePatcher. I'd say pull this one in first and then I can resolve conflicts on mine

ChaoticTempest:
sounds good then

PR #75, opened 2022-02-09
feat: add rpc logging

itegulov:
This PR adds logging to RPC client. I made the logs dynamic depending on the logging level enabled. Usually `INFO` would be enough for most users as it prints one-liners with submitted actions and the resulting status. But if a user wants more information (e.g. gas profile, logs etc), they can enable `DEBUG`. Try it out by running tests with `RUST_LOG=workspaces=info`.

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -38,17 +38,6 @@ impl Client {
         Self { rpc_addr }
     }
 
-    async fn query_retry<M: methods::RpcMethod, F, T>(
-        &self,
-        f: F,
-    ) -> MethodCallResult<M::Response, M::Error>
-    where
-        F: Fn() -> T,
-        T: core::future::Future<Output = MethodCallResult<M::Response, M::Error>>,
-    {
-        retry(|| async { f().await }).await
-    }
-
     pub(crate) async fn query_broadcast_tx(
         &self,
         method: &methods::broadcast_tx_commit::RpcBroadcastTxCommitRequest,
@@ -56,10 +45,13 @@ impl Client {
         FinalExecutionOutcomeView,
         near_jsonrpc_primitives::types::transactions::RpcTransactionError,
     > {
-        self.query_retry::<methods::broadcast_tx_commit::RpcBroadcastTxCommitRequest, _, _>(|| async {
+        retry(|| async {
             let result = JsonRpcClient::connect(&self.rpc_addr).call(method).await;
             match &result {
                 Ok(response) => {
+                    // When user sets logging level to INFO we only print one-liners with submitted
+                    // actions and the resulting status. If the level is DEBUG or lower, we print
+                    // the entire request and response structures.
                     if tracing::level_enabled!(tracing::Level::DEBUG) {
                         tracing::debug!(
                             target: "workspaces",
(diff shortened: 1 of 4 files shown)


workspaces/src/rpc/client.rs
```diff
@@ -37,11 +38,84 @@ impl Client {
...
+        F: Fn() -> T,
+        T: core::future::Future<Output = MethodCallResult<M::Response, M::Error>>,
+    {
+        retry(|| async { f().await }).await
+    }
+
+    pub(crate) async fn query_broadcast_tx(
+        &self,
+        method: &methods::broadcast_tx_commit::RpcBroadcastTxCommitRequest,
+    ) -> MethodCallResult<
```
itegulov, line 55: I don't really like how I had to create a separate function for broadcast txs, but it seems like Rust does not support specialization, so I do not know if there is a better way... Any suggestions are welcome

workspaces/src/rpc/client.rs
```diff
@@ -37,11 +38,84 @@ impl Client {
...
+        FinalExecutionOutcomeView,
+        near_jsonrpc_primitives::types::transactions::RpcTransactionError,
+    > {
+        self.query_retry::<methods::broadcast_tx_commit::RpcBroadcastTxCommitRequest, _, _>(|| async {
```
ChaoticTempest, line 59: since `retry(|| async { f().await }).await` can just be `retry(f).await`, which is like a thin wrapper that doesn't do much, we can just get rid of `query_retry` and do:
```suggestion
        retry(|| async {
```

workspaces/src/rpc/client.rs
ChaoticTempest: yeah there's not much you can do here besides this. The alternative is creating some traits to specialize over, but that's not gonna be great either

workspaces/src/rpc/client.rs
```diff
@@ -37,11 +38,84 @@ impl Client {
...
+            let result = JsonRpcClient::connect(&self.rpc_addr).call(method).await;
+            match &result {
+                Ok(response) => {
+                    if tracing::level_enabled!(tracing::Level::DEBUG) {
```
ChaoticTempest, line 63: probably want to add some comments here why we're doing two different log messages

workspaces/src/rpc/client.rs
```diff
@@ -37,11 +38,84 @@ impl Client {
...
+                            error
+                        );
+                    } else {
+                        tracing::error!(
+                            target: "workspaces",
+                            "Submitting transaction with actions {:?} resulted in error {:?}",
+                            method.signed_transaction.transaction.actions,
+                            error
+                        );
+                    }
```
ChaoticTempest, line 94: hmm, I'm debating this, but for errors, we should be as verbose as possible anyways. There could be cases where errors are hard to reproduce

ChaoticTempest (commented):
looking good so far and adds a lot of value to how we can debug errors in workspaces. Only a couple things I noticed we can make simpler

workspaces/src/rpc/client.rs
itegulov: Yeah, good point, this common function is pretty useless

workspaces/src/rpc/client.rs
itegulov: You are right. Since errors should not happen very often and they could be hard to reproduce I think it makes sense to print them verbosely all the time.

PR #77, opened 2022-02-18
feat: add a way to automatically compile and deploy a contract

### LINKED ISSUE #54: Figure out an easy way to both compile contracts and run them with workspaces
This might be a tricky one to solve, but ideally, we should have a way that allows the users of the both the sdk and workspaces to compile a contract, and then automatically point it towards workspaces for it to be deployed. We could scrape the local project's `Cargo.toml` for this info, but there be dragons if it were nested into multiple members.

```rust
worker.dev_deploy(workspaces::this_project());
```
or if multiple contracts
```rust
worker.dev_deploy(workspaces::this_project("ref-finance"));
```
*syntax not final

itegulov:
Fixes #54 

This is a fairly naive implementation that I have tested on all examples from near-sdk-rs. So it should still suit the majority of users, but I can see the following paths for further improvement:
* Make `dev_deploy_project` more flexible by introducing some sort of builder where a user would be able to pass custom cargo arguments (or even switch between `debug` and `release` as I am just using `release` by default for everything), env variables, target wasm filename if it is ambiguous (I am not sure if a single project can generate >1 wasm files, but it probably can) etc
* Introduce a build cache with a mutex for each unique project path. Build cache would save the generated `.wasm` file to `/tmp` and all other tests will use it instead of trying to re-compile the project. This will be especially useful for people who have long-running tests that want to edit their codebase while the tests are running.
* Not sure if this is possible/good idea, but maybe a macro that would do contract compilation at test compile time and save the wasm contents directly to the test binary?
* Low priority, but would be cool to set up integration tests that would try to compile & deploy some well-known NEAR contracts (kind of like compiler devs do it with third-party libraries to prevent regression)


workspaces/src/cargo/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+use cargo_metadata::Message;
+use std::fmt::Debug;
+use std::path::Path;
+use std::process::{Command, Stdio};
+use tracing::debug;
+
+pub fn build_cargo_project<P: AsRef<Path> + Debug>(
+    project_path: P,
+) -> anyhow::Result<Vec<Message>> {
+    let cmd = Command::new("cargo")
```
ChaoticTempest, line 11: have you tried making use of the async version of `Command` (https://crates.io/crates/async-process)? That would probably be better to use since we're always gonna be in an async context. Starting up sandbox should switch over to that as well

workspaces/src/cargo/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+            "--target",
+            "wasm32-unknown-unknown",
+            "--release",
+            "--message-format=json",
+        ])
+        .current_dir(&project_path)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()?;
+    let output = cmd.wait_with_output()?;
```
ChaoticTempest, line 23: and consequently we get rid of this block

ChaoticTempest (commented):
Overall looks good. Some stuff that might be relevant to model off of since dtolnay has already done custom cargo commands: https://github.com/dtolnay/trybuild/blob/master/src/cargo.rs

> I am just using release by default for everything), env variables, target wasm filename if it is ambiguous...

release is fine since for the most part, the sdk is heavily optimized, and we'd want to test for the release binary anyways.

> (I am not sure if a single project can generate >1 wasm files, but it probably can) etc

It can if we end up specifying the path to a `cargo-workspace` project which specifies a bunch of workspace members. But yeah, we can leave behind compiling multiple contracts and lets just force users to specify the specific contract path instead then. This would makes things a lot simpler.

> Introduce a build cache with a mutex for each unique project path. Build cache would save the generated .wasm file to /tmp and all other tests will use it instead of trying to re-compile the project. This will be especially useful for people who have long-running tests that want to edit their codebase while the tests are running.

hmm, since we're running the cargo command, cargo would lock further compilation on the same project wouldn't it? Like say test1 already compiles, test2 tries to compiles but there's no changes so cargo wouldn't be triggered at all. Similarly, if they weren't sequential tests, there's a Cargo lockfile that blocks test2 from running cargo. That's what I think happens at least, so might not need need this build cache

> would be cool to set up integration tests that would try to compile & deploy some well-known NEAR contracts (kind of like compiler devs do it with third-party libraries to prevent regression)

would definitely like to have some more well known contracts tested. We have a ref-finance contract in the examples, but we don't run github actions on examples as of now

itegulov:
@ChaoticTempest 
> Some stuff that might be relevant to model off of since dtolnay has already done custom cargo commands: [https://github.com/dtolnay/trybuild/blob/master/src/cargo.rs](https://github.com/dtolnay/trybuild/blob/master/src/cargo.rs?rgh-link-date=2022-02-25T00%3A49%3A00Z)

Thanks for the pointer, I will take a look

> release is fine since for the most part, the sdk is heavily optimized, and we'd want to test for the release binary anyways.

Yeah, that is why I chose it to be the default. I was just thinking that there _could_ be some use cases where a user would want to use custom cargo commands that I am not aware of, but I guess we can gather some feedback from users before stabilizing this feature. WDYT in general about making this an `unstable` feature for now?

> hmm, since we're running the cargo command, cargo would lock further compilation on the same project wouldn't it? Like say test1 already compiles, test2 tries to compiles but there's no changes so cargo wouldn't be triggered at all. Similarly, if they weren't sequential tests, there's a Cargo lockfile that blocks test2 from running cargo. That's what I think happens at least, so might not need need this build cache

I think you misunderstood my point here. Since we are trying to compile codebase located on the user's machine it can change at any time during the test execution. Let's say I am running 16 long-running tests in parallel using 8 threads. As you said, test 1 will successfully build the cargo project and tests 2-8 will try to build but since there will be nothing to do and they will just reuse the existing compilation artifact. But each of these tests requires 20 minutes to finish, in the meantime an impatient developer starts making changes to the contract codebase. Tests 1-8 finish and now it's turn for tests 9-16; they invoke `cargo build` but it fails because the project is already in an incompilable state because the user was in the middle of writing a new function. Or, even worse, it was in a compilable state but tests 9-16 used a different wasm file that has a freshly introduced bug. Imagine how confusing it would be :sweat_smile:

ChaoticTempest:
> WDYT in general about making this an unstable feature for now?

That's a good idea. It's definitely more of something that could help us, but the experience just isn't fully hammered out. Let's leave it for experimental reasons then

> because the user was in the middle of writing a new function

Gotcha. Yeah then all the more reason to make this feature unstable. It would definitely be unintuitive, but if its under a feature flag, we can educate users on the caveats of this feature if we don't immediately go the build cache route

workspaces/src/cargo/mod.rs
```diff
@@ -0,0 +1,46 @@
+use anyhow::anyhow;
+use cargo_metadata::Message;
+use std::fmt::Debug;
+use std::path::Path;
+use std::process::{Command, Stdio};
+use tracing::debug;
+
+pub fn build_cargo_project<P: AsRef<Path> + Debug>(
+    project_path: P,
+) -> anyhow::Result<Vec<Message>> {
```
austinabell, line 10: prob not ideal to expose `cargo_metadata` types in our API

workspaces/src/network/mod.rs
```diff
@@ -58,6 +60,13 @@ pub trait DevAccountDeployer {
     async fn dev_generate(&self) -> (AccountId, SecretKey);
     async fn dev_create_account(&self) -> anyhow::Result<Account>;
     async fn dev_deploy(&self, wasm: &[u8]) -> anyhow::Result<Contract>;
+
+    /// Builds the cargo project located at `project_path` and deploys the resulting smart
+    /// contract using a freshly generated dev account.
+    ///
+    /// NOTE: This function does not check whether the resulting wasm file is a valid smart
+    /// contract or not.
+    async fn dev_deploy_project(&self, project_path: &str) -> anyhow::Result<Contract>;
```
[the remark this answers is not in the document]
austinabell: Seems like it probably shouldn't have anything to do with the network-related trait. There are no ties to the network for this

workspaces/src/network/mod.rs
```diff
@@ -92,6 +101,40 @@ where
...
+            .filenames
+            .iter()
+            .filter(|f| f.as_str().ends_with(".wasm"))
+            .collect::<Vec<_>>();
+        match wasm_files.as_slice() {
+            [] => Err(anyhow!(
+                "Compilation resulted in no '.wasm' target files. \
+                 Please check that your project contains a NEAR smart contract."
+            )),
+            [file] => self.dev_deploy(&fs::read(file.canonicalize()?)?).await,
```
austinabell, line 131: the file reading could be async here, and my intuition is that it's probably best to, although probably doesn't matter in most cases

workspaces/tests/test-contracts/ft/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "test-contract-ft"
+version = "0.0.0"
+authors = ["Near Inc <hello@nearprotocol.com>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
austinabell, line 8:
```suggestion
crate-type = ["cdylib"]
```

workspaces/tests/test-contracts/ft/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+    env, log, near_bindgen, require, AccountId, Balance, BorshStorageKey, PanicOnDefault,
+    PromiseOrValue,
+};
+
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct Contract {
+    token: FungibleToken,
+    metadata: LazyOption<FungibleTokenMetadata>,
+}
```
austinabell, line 18: Why is FT contract used specifically? Seems like you're just trying to make sure that it can be compiled and called in a test? Seems like having a more involved test like this would just be more annoying for maintenance unless there is a reason for it?

workspaces/src/network/mod.rs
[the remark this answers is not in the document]
itegulov: That's fair, I moved all logic to cargo module

workspaces/src/cargo/mod.rs
itegulov: Done!

workspaces/src/cargo/mod.rs
itegulov: Yeah, I should have made this `pub(crate)` but since I moved all compilation logic into `cargo`, this can be private now

workspaces/tests/test-contracts/ft/src/lib.rs
itegulov: Good point, replaced with `status-message` which is much simpler.

workspaces/tests/deploy_project.rs
```diff
@@ -0,0 +1,26 @@
+#![recursion_limit = "256"]
```
ChaoticTempest, line 1: We'll also want `#![cfg(feature = "unstable")]` on top since `cargo test` will try to run it

itegulov: right, I have also made so that GHA runs tests with `--features unstable` since that is what we are doing in near-sdk-rs, but lmk if you want me to create two separate workflows for stable/unstable version

.github/workflows/test.yml
```diff
@@ -38,6 +38,6 @@ jobs:
     - name: Add wasm32 target
       run: rustup target add wasm32-unknown-unknown
     - name: Build
-      run: cargo build --verbose
+      run: cargo build --verbose --features unstable
```
austinabell, line 41: not sure why this is `build`, can just be `cargo check`, also prob worth checking without unstable feature in this workflow, should be low additional cost

workspaces/src/cargo/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::fmt::Debug;
+use std::fs;
+use std::path::Path;
+use std::process::Stdio;
+use tracing::debug;
+
+async fn build_cargo_project<P: AsRef<Path> + Debug>(
+    project_path: P,
+) -> anyhow::Result<Vec<Message>> {
+    let output = Command::new("cargo")
```
miraclx, line 13: What do you think about reusing the target directory of the workspace instead of having isolated spaces for each contract.

`cargo build` has `--target-dir` and `cargo_metadata::Metadata` has [`target_directory`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#structfield.target_directory)

Might be worth investigating this kind of behavior as target directories tend to take up a lot of disk space, having reusable artifacts would be a +1.

itegulov: Hmm, I am not sure I understand your point. Are you suggesting we reuse the same `workspaces` artifact for multiple contracts? But an average user wouldn't have a local copy of `workspaces` on their machine to share, would they? But using `--target-dir` to point cargo to a non-default directory might actually be worthwhile to avoid collision with user-generated compilation artifacts.

itegulov:
@ChaoticTempest 
> Some stuff that might be relevant to model off of since dtolnay has already done custom cargo commands: [https://github.com/dtolnay/trybuild/blob/master/src/cargo.rs](https://github.com/dtolnay/trybuild/blob/master/src/cargo.rs?rgh-link-date=2022-02-25T07%3A14%3A28Z)

I have pulled a configurable cargo binary from there. Also tried out offline mode, but I misunderstood how it works and despite it claiming that it **avoids** making network calls it actually outright refuses to download dependencies if you don't already have them which isn't good.

workspaces/src/cargo/mod.rs
miraclx: Oh, to be clear, by "workspace" I didn't mean `workspaces-rs`, but instead `cargo` "workspace". I meant, reusing the target directory of the environment compiling the contract. So all the build artifacts resides in one place.

Following the current logic, `workspaces-rs` would cd into the contract directory and run `cargo build` in there, creating a separate `target` directory at that tree depth.

My suggestion was to research potentially merging the target directories. So contracts have their artifacts in the same `target-dir` as the project compiling them. The point of the research is see how well cargo plays with having multiple projects in the same space. My gut feeling is it would behave the same as a normal cargo workspace with members. But I'm not certain 🤞🏽.

itegulov: Ah, yeah, now I understand what you mean. Indeed, this seems like a good idea, let me give it a try.

itegulov: Works like a charm! Thanks for the suggestion

workspaces/src/cargo/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+fn cargo_bin() -> Command {
+    match env::var_os("CARGO") {
+        Some(cargo) => Command::new(cargo),
+        None => Command::new("cargo"),
+    }
+}
+
+fn cargo_metadata<P: AsRef<Path> + Debug>(project_path: P) -> anyhow::Result<Metadata> {
+    let mut metadata_command = MetadataCommand::new();
+    metadata_command.current_dir(project_path.as_ref());
```
miraclx, line 20: @itegulov, so this actually wasn't what I meant either 😅.. How about you try omitting this line and not setting `current_dir`, so when `cargo` builds the contract, it'll reuse the target directory of the project. Instead of creating a separate one inside the contract directory.

.gitignore
```diff
@@ -1,5 +1,6 @@
 # Build files
 /target
+**/target/
```
miraclx, line 3: ^^ If that works, for example, we wouldn't need this line.

workspaces/src/cargo/mod.rs
itegulov: Well, `cargo metadata` should not build anything anyway, right? If I get rid of `current_dir`, then how would `CargoCommmand` know where the directory containing the contract is located at?

miraclx: `cargo build` already knows that from here: https://github.com/near/workspaces-rs/blob/eff8274fc35418b668d59b29887b601b2112f60f/workspaces/src/cargo/mod.rs#L38

The point of not setting the `current_dir` for `cargo metadata` is so we can get the `target` directory of the parent.

So, while `cargo` builds in the directory containing the contract, it keeps all the build artifacts in the same `target` directory.

miraclx: For example, in the case of the `test-contract-status-message` contract, we have a tree like this:

```diff
 near/workspaces-rs
 ├── target
+│   └── wasm32-unknown-unknown
+│       └── release
+│           └── test_contract_status_message.wasm
 └── workspaces
     └── tests
         └── test-contracts
             └── status-message
-                └── target
-                    └── wasm32-unknown-unknown
-                        └── release
-                            └── test_contract_status_message.wasm
```

itegulov: > The point of not setting the current_dir for cargo metadata is so we can get the target directory of the parent.

Exactly, when you do `cargo metadata` in one of the workspaces, it returns `target_directory` that is equal to the target directory of the parent.

> So, while cargo builds in the directory containing the contract, it keeps all the build artifacts in the same target directory.

And to ensure that it works as you describe it I set `--target-dir` to the output of `cargo metadata` from the previous paragraph.

Sorry if I am speaking nonsense here, but I am really confused how is my approach is different to what you are suggesting 😅

itegulov: Hmm, but `test-contract-status-message` is not a workspace of the root cargo project, but rather an independent cargo project. Would it be safe to reuse root target directory for building it?

ChaoticTempest: I'd say its pretty safe to put it in the root target dir. Pretty much everything workspaces-rs generates goes in there including the downloaded sandbox binary. It's also a very clean way of later deleting everything with just a `cargo clean` at the root level

itegulov: Ok, I have tested this and indeed it seems to work fine. I could find any definitive information on "safeness" to do so, but I guess since the feature is unstable anyway we can try and see if it works as expected for users?

workspaces/tests/deploy_project.rs
```diff
@@ -0,0 +1,27 @@
+#![cfg(feature = "unstable")]
+#![recursion_limit = "256"]
+use test_log::test;
+use workspaces::prelude::*;
+
+#[test(tokio::test)]
+async fn test_dev_deploy_project() -> anyhow::Result<()> {
+    let worker = workspaces::sandbox();
+    let wasm = workspaces::compile_project("./tests/test-contracts/status-message").await?;
```
miraclx, line 9: TBH, this path is a bit confusing.. it's not immediately clear that it's relative to the `workspaces` folder (it reads as though this path points to `workspaces/tests/tests/test-contracts/status-message`).

Ideally, given the location of this file, shouldn't the contract path be `./test-contracts/status-message`?

I'm thinking maybe we should wrap the `compile_project` function in a macro like `compile!()` that internally checks `file!()`, getting the parent and resolving relative paths from there (leaving paths as-is if they're already absolute).

Something like this:

```rust
#[macro_export]
macro_rules! compile {
    ($contract_path:expr) => {
        $crate::__compile_contract(::std::path::Path::new(file!()), $contract_path)
    };
}
```

Although there's an issue when using a cargo workspace with members (as with the case of `workspaces-rs`), because `file!()` returns the path relative to the root workspace and not the member crate itself. So, from `deploy_project.rs`, `parent(file!())` returns `workspaces/tests`

I'm thinking we could use `cargo-metadata` to first, return all the members, with their respective paths, we can then find the member that owns the path returned by `parent(file!())` and resolve the relative path from there.

Although, you'd have to introduce new dependencies like [`pathdiff`](https://docs.rs/pathdiff/0.2.1/pathdiff/fn.diff_paths.html) to resolve relative paths, seeing as the stdlib doesn't have that functionality.

miraclx: Resolved in https://github.com/near/workspaces-rs/pull/89

miraclx (approved):

austinabell (approved):

workspaces/Cargo.toml
```diff
@@ -10,9 +10,11 @@ Library for automating workflows and testing NEAR smart contracts.
 
 [dependencies]
 async-trait = "0.1"
+async-process = "1.3.0"
```
ChaoticTempest, line 13: might be a good idea to make this optional under the `unstable` flag as well, but not a big deal

workspaces/tests/deploy_project.rs
```diff
@@ -0,0 +1,27 @@
...
+    let contract = worker.dev_deploy(&wasm).await?;
+
+    let _res = contract
+        .call(&worker, "set_status")
+        .args_json(("foo",))?
+        .gas(300_000_000_000_000)
```
ChaoticTempest, line 15: could replace this to `.max_gas()` now that https://github.com/near/workspaces-rs/pull/94 has made it in

ChaoticTempest (approved):

workspaces/Cargo.toml
austinabell: Good catch! Yeah, it probably should be optional and pulled in with this feature

PR #79, opened 2022-03-01
refactor: make patch state key `Vec<u8>` instead of `String`

### LINKED ISSUE #51: patch state key should be `Vec<u8>` and not `String`
Currently only usage uses string keys, but this should not be required to be utf8 bytes

itegulov:
Fixes #51

### DIFF
--- a/workspaces/src/network/mod.rs
+++ b/workspaces/src/network/mod.rs
@@ -101,7 +101,7 @@ pub trait StatePatcher {
     async fn patch_state(
         &self,
         contract_id: &AccountId,
-        key: String,
+        key: Vec<u8>,
         value: Vec<u8>,
     ) -> anyhow::Result<()>;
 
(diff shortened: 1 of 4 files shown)


ChaoticTempest (approved):

workspaces/src/network/mod.rs
```diff
@@ -120,12 +120,12 @@ where
     async fn patch_state(
         &self,
         contract_id: &AccountId,
-        key: String,
+        key: Vec<u8>,
         value: Vec<u8>,
```
austinabell, line 124: these should probably be slices to be consistent with the rest of the API, or is there a reason they are not @ChaoticTempest ?

ChaoticTempest: ahh, yeah best to keep them as slices. My bad, prematurely merged this one since it looked simple

itegulov: no worries, I can fix this real quick, I just went by the issue description

you want me to change both key and value to be [u8], right?

ChaoticTempest: yup both should be slices

austinabell: was unrelated to this PR, just wanted to point out before opening an issue :)

PR #80, opened 2022-03-01
refactor: change patch state parameters to be slices

itegulov:
In the end, I had to copy the slices into vectors anyway, not sure if there is a way to avoid that

### DIFF
--- a/examples/src/spooning.rs
+++ b/examples/src/spooning.rs
@@ -99,8 +99,8 @@ async fn main() -> anyhow::Result<()> {
     worker
         .patch_state(
             sandbox_contract.id(),
-            "STATE".into(),
-            status_msg.try_to_vec()?,
+            "STATE".as_bytes(),
+            &status_msg.try_to_vec()?,
         )
         .await?;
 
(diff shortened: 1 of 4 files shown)


ChaoticTempest (approved):
Yeah that's the unfortunate part about these APIs. It's to keep them as convenient as possible for users to make use of, which means not having the most performant code underneath, which is fine in the context of workspaces

adsick:
`Into<Vec<u8>>` may work, I already try that:
path_state.rs:
```rust
    worker
        .patch_state(&contract_id, "STATE", status_msg.try_to_vec()?)
        .await?;
```
worker/impls.rs
```rust
    pub async fn patch_state(
        &self,
        contract_id: &AccountId,
        key: impl Into<Vec<u8>>,
        value: impl Into<Vec<u8>>,
    ) -> anyhow::Result<()> {
        self.workspace.patch_state(contract_id, key, value).await
    }
```
network/sandbox.rs
```rust
    pub(crate) async fn patch_state(
        &self,
        contract_id: &AccountId,
        key: impl Into<Vec<u8>>,
        value: impl Into<Vec<u8>>,
    ) -> anyhow::Result<()> {
        let state = StateRecord::Data {
            account_id: contract_id.to_owned(),
            data_key: key.into(),
            value: value.into(),
        };
...
```

Issue #76, opened 2022-02-17
Do not have access on account's balance/block_timestamp


Buckram123:
I feel like adding a method like `view_account()` to Worker to check account balance would be great. Also a way to check block_timestamp might be useful

PR #82, opened 2022-03-02
chore: Add view_account, view_code, view_block

ChaoticTempest:
This adds a couple convenience functions to `Worker` and `Account` types:
- view_account: stuff like balance and storage usage
- view_code
- view_block: to grab latest block timestamps

Also addresses https://github.com/near/workspaces-rs/issues/76

### DIFF
--- a/workspaces/src/network/account.rs
+++ b/workspaces/src/network/account.rs
@@ -1,10 +1,11 @@
 use std::convert::TryInto;
 
 use near_crypto::KeyType;
+use near_primitives::views::AccountView;
 
 use crate::rpc::client::{DEFAULT_CALL_DEPOSIT, DEFAULT_CALL_FN_GAS};
 use crate::types::{AccountId, Balance, Gas, InMemorySigner, SecretKey};
-use crate::{Network, Worker};
+use crate::{CryptoHash, Network, Worker};
 
 use super::{CallExecution, CallExecutionDetails, ViewResultDetails};
 
@@ -73,6 +74,11 @@ impl Account {
         worker
             .delete_account(&self.id, &self.signer, beneficiary_id)
             .await
+    }
+
+    /// Views the current account's details such as balance and storage usage.
+    pub async fn details<T: Network>(&self, worker: &Worker<T>) -> anyhow::Result<AccountDetails> {
+        worker.view_account(&self.id).await
     }
 
     /// Create a new sub account. Returns a CreateAccountBuilder object that
@@ -163,6 +169,11 @@ impl Contract {
         args: Vec<u8>,
     ) -> anyhow::Result<ViewResultDetails> {
         worker.view(self.id(), function, args).await
+    }
+
+    /// Views the current contract's details such as balance and storage usage.
+    pub async fn details<T: Network>(&self, worker: &Worker<T>) -> anyhow::Result<AccountDetails> {
+        worker.view_account(self.id()).await
     }
 
     /// Deletes the current contract, and returns the execution details of this
@@ -312,5 +323,26 @@ where
             result: account,
             details: outcome.into(),
         })
+    }
+}
+
+/// Details of an Account or Contract. This is an non-exhaustive list of items
+/// that the account stores in the blockchain state.
+#[non_exhaustive]
+pub struct AccountDetails {
+    pub amount: Balance,
+    pub locked: Balance,
+    pub code_hash: CryptoHash,
+    pub storage_usage: u64,
+}
+
+impl From<AccountView> for AccountDetails {
+    fn from(account: AccountView) -> Self {
+        Self {
+            amount: account.amount,
+            locked: account.locked,
+            code_hash: CryptoHash(account.code_hash.0),
+            storage_usage: account.storage_usage,
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


workspaces/src/network/block.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// The block header info. This is a non-exhaustive list of items that
+/// could be present in a block header. More can be added in the future.
+#[non_exhaustive]
+struct BlockHeader {
+    height: BlockHeight,
+    epoch_id: CryptoHash,
+    hash: CryptoHash,
+    timestamp_nanosec: u64,
+}
```
itegulov, line 40: How did you judge which items to include here, user feedback? Not that I am against it, just wondering

workspaces/src/worker/impls.rs
```diff
@@ -127,6 +133,10 @@ where
         self.client().view_state(contract_id.clone(), prefix).await
     }
 
+    pub async fn view_latest_block(&self) -> anyhow::Result<Block> {
+        self.client().view_block(None).await.map(Into::into)
+    }
```
itegulov, line 138: nit: seems like you forgot documentation for this method

workspaces/src/worker/impls.rs
```diff
@@ -150,6 +160,13 @@ where
             .await
             .map(Into::into)
     }
+
+    pub async fn view_account(&self, account_id: &AccountId) -> anyhow::Result<AccountDetails> {
+        self.client()
+            .view_account(account_id.clone(), None)
+            .await
+            .map(Into::into)
+    }
```
itegulov, line 169: Same here

workspaces/src/worker/impls.rs
itegulov, line 138: Hmm, does this mean there will be no way to view an arbitrary (non-last) block?

workspaces/src/network/block.rs
ChaoticTempest: ahh I forgot to put this in the initial PR post. The fields I grabbed from the block header was what seemed important for now and was thinking of exposing more later. Would you have any thoughts of including them all now?

workspaces/src/worker/impls.rs
ChaoticTempest: yup, and because it would only be available for archival networks. The archival stuff in workspaces is very limited to just working with importing contracts

itegulov: Ah, ok, I was confused because the underlying client function you implemented supports any kind of block.

workspaces/src/network/block.rs
itegulov: Nah, that's probably good enough for now. We can always add more, taking stuff out will be more difficult

itegulov (approved):

workspaces/src/network/account.rs
```diff
@@ -314,3 +336,24 @@ where
         })
     }
 }
+
+/// Details of an Account or Contract. This is an non-exhaustive list of items
+/// that the account stores in the blockchain state.
+#[non_exhaustive]
+pub struct AccountDetails {
+    pub amount: Balance,
```
austinabell, line 344: should this not be named balance? Amount seems very ambiguous

workspaces/src/network/account.rs
```diff
@@ -314,3 +336,24 @@ where
         })
     }
 }
+
+/// Details of an Account or Contract. This is an non-exhaustive list of items
+/// that the account stores in the blockchain state.
+#[non_exhaustive]
+pub struct AccountDetails {
```
austinabell, line 343: can you add Debug and PartialEq at least to all these added types?

workspaces/src/network/block.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// The id of an epoch this block belongs to.
+    pub fn epoch_id(&self) -> &CryptoHash {
+        &self.header.epoch_id
+    }
+}
+
+/// The block header info. This is a non-exhaustive list of items that
+/// could be present in a block header. More can be added in the future.
+#[non_exhaustive]
```
austinabell, line 34: noting this tag doesn't really matter since it's a private type

workspaces/src/types.rs
```diff
@@ -93,3 +96,15 @@ impl TryFrom<Vec<u8>> for CryptoHash {
         <Self as TryFrom<&[u8]>>::try_from(v.as_ref())
     }
 }
+
+impl fmt::Debug for CryptoHash {
```
miraclx, line 100: `impl std::hash::Hash` seems like a reasonable addition.

miraclx (approved):

workspaces/src/types.rs
austinabell: Why would you want to hash a hash?

austinabell (approved):

workspaces/src/types.rs
miraclx: So you can use it as keys to `HashMap` & `HashSet`

austinabell: Yeah, could derive it, probably worth adding

austinabell: > So you can use it as keys to `HashMap` & `HashSet`

oops had deleted my comment and updated it before I saw you responded. Sorry for my brain lapse :D

PR #83, opened 2022-03-03
feat: improve error handling on failed transactions

ChaoticTempest:
This PR improves error handling on failed transactions and will force the user to handle the error now as an anyhow result. For instance, before this PR, the following would just dangle with the status object returned being ignored at the end.
```
contract.call(function_name)
   .args(...)
   .transact()
   .await?;
```
And after this PR, the `?` error forward at the end will trigger an error on failed transactions. This improves propagating errors such as invalid arguments to calling into a contract function

### DIFF
--- a/workspaces/src/network/result.rs
+++ b/workspaces/src/network/result.rs
@@ -81,6 +81,22 @@ impl CallExecutionDetails {
         };
         base64::decode(result).map_err(Into::into)
     }
+
+    /// Convert the execution details into a Result if its status is not a successful one.
+    /// Useful for checking if the call was successful and forwarding the error upwards.
+    fn try_into_result(self) -> anyhow::Result<Self> {
+        match self.status {
+            FinalExecutionStatus::Failure(ref err) => anyhow::bail!(err.clone()),
+            FinalExecutionStatus::NotStarted => anyhow::bail!("Transaction not started."),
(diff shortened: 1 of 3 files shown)


itegulov (approved):
Makes total sense to me

workspaces/src/network/result.rs
```diff
@@ -81,6 +81,22 @@ impl CallExecutionDetails {
...
         base64::decode(result).map_err(Into::into)
     }
+
+    /// Convert the execution details into a Result if its status is not a successful one.
+    /// Useful for checking if the call was successful and forwarding the error upwards.
+    fn try_into_result(self) -> anyhow::Result<Self> {
+        match self.status {
+            FinalExecutionStatus::Failure(ref err) => anyhow::bail!(err.clone()),
+            FinalExecutionStatus::NotStarted => anyhow::bail!("Transaction not started."),
+            FinalExecutionStatus::Started => anyhow::bail!("Transaction still being processed."),
```
austinabell, line 91: noting that a user might want to use the details of the call execution when handling the error, even if it's downcasting to the type.

Fine as is for now though I think

austinabell (approved):

workspaces/src/network/result.rs
ChaoticTempest: yeah we should make a custom Error to capture some of the details like logs later when the other refactor/results PR lands

PR #84, opened 2022-03-03
fix: make empty array the default function argument

### LINKED ISSUE #36: Investigate inner JSON RPC serialization
Currently, default args supplied to calling into a contract change function is an empty json map `{}`. This is required when we want to supply no arguments to a change function that take all `Option`als. This might just be a weird case in how jsonrpc is interpreting/serializing the bytes underneath.

Additionally, functions that take no arguments are fine with either empty bytes (`Vec::new()`) or the empty json map (`json!({})`).

We'll need to take a look closer to see whether this is the right thing to be using or not. But for now, `json({})` will be the default.

Original discussion of this here: https://github.com/near/workspaces-rs/pull/33#discussion_r768266227_

itegulov:
Fixes #36

### DIFF
--- a/workspaces/src/network/account.rs
+++ b/workspaces/src/network/account.rs
@@ -199,7 +199,7 @@ impl<'a, T: Network> CallBuilder<'a, T> {
             signer,
             contract_id,
             function,
-            args: serde_json::json!({}).to_string().into_bytes(),
+            args: vec![],
             deposit: DEFAULT_CALL_DEPOSIT,
             gas: DEFAULT_CALL_FN_GAS,
         }


ChaoticTempest (approved):

ChaoticTempest:
@itegulov I forgot, there's one thing we should correct before pulling this in. The ref_finance example has some calls into function with optionals, which will error out with this PR. Can you fix that when you get the chance. We should be able to see which one fails with https://github.com/near/workspaces-rs/pull/83 being pulled in

itegulov:
@ChaoticTempest indeed, should be fixed now. Do you want me to convert these examples to be tests and add them to CI in a separate PR?

ChaoticTempest:
it be cool if we had a test that tests for optional args specifically, so don't need to pull all of the example in to tests

itegulov:
@ChaoticTempest added. PTAL when you can.

workspaces/tests/optional_args.rs
```diff
@@ -0,0 +1,57 @@
...
+    let worker = workspaces::sandbox();
+    let contract = init(&worker).await?;
+
+    let res = contract
+        .call(&worker, "storage_unregister")
+        .gas(300_000_000_000_000)
+        .deposit(1)
+        .transact()
+        .await;
+    assert!(format!("{:?}", res.unwrap_err()).contains("Failed to deserialize input from JSON"));
```
ChaoticTempest, line 35: yeah this error message is pretty unintuitive, so best for us to change this on the SDK side in the future

ChaoticTempest (approved):

PR #85, opened 2022-03-04
chore: suppress sandbox bin logs

ChaoticTempest:
This suppresses all sandbox logs and allows the user to control it via an environment variable

### DIFF
--- a/workspaces/src/network/server.rs
+++ b/workspaces/src/network/server.rs
@@ -77,10 +77,9 @@ impl Drop for SandboxServer {
 /// will be forward into RUST_LOG environment variable as to not conflict
 /// with similar named log targets.
 fn supress_sandbox_logs_if_required() {
-    match std::env::var("NEAR_ENABLE_SANDBOX_LOG") {
-        Ok(val) if val == "on" => {}
+    if let Err(_) = std::env::var("NEAR_ENABLE_SANDBOX_LOG") {
         // non-exhaustive list of targets to supress, since choosing a default LogLevel
         // does nothing in this case, since nearcore seems to be overriding it somehow:
-        _ => std::env::set_var("NEAR_SANDBOX_LOG", "near=error,stats=error,network=error"),
+        std::env::set_var("NEAR_SANDBOX_LOG", "near=error,stats=error,network=error");
     };
 }
--- a/workspaces/src/network/server.rs
+++ b/workspaces/src/network/server.rs
@@ -72,14 +72,18 @@ impl Drop for SandboxServer {
 }
 
 /// Turn off neard-sandbox logs by default. Users can turn them back on with
-/// NEAR_ENABLE_SANDBOX_LOG=on and specify further paramters with the custom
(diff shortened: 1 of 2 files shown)


workspaces/src/network/server.rs
```diff
@@ -67,3 +70,17 @@ impl Drop for SandboxServer {
...
 }
+
+/// Turn off neard-sandbox logs by default. Users can turn them back on with
+/// NEAR_ENABLE_SANDBOX_LOG=on and specify further paramters with the custom
+/// NEAR_SANDBOX_LOG for higher levels of specificity. NEAR_SANDBOX_LOG args
+/// will be forward into RUST_LOG environment variable as to not conflict
+/// with similar named log targets.
+fn supress_sandbox_logs_if_required() {
+    match std::env::var("NEAR_ENABLE_SANDBOX_LOG") {
+        Ok(val) if val == "on" => {}
```
austinabell, line 81: why not just require the env variable to be set? is the `on` pattern documented somewhere?

austinabell (approved):
didn't functionally test but lgtm

workspaces/src/network/server.rs
itegulov: I agree, the usual convention is either set/unset or 1/0

itegulov (approved):
Tested, works fine

workspaces/src/network/server.rs
ChaoticTempest: Ahh, will switch to just set/unset. Way too use to `cmake` flags so a force of habit

workspaces/src/network/server.rs
```diff
@@ -67,3 +70,16 @@ impl Drop for SandboxServer {
             .unwrap();
     }
 }
+
+/// Turn off neard-sandbox logs by default. Users can turn them back on with
+/// NEAR_ENABLE_SANDBOX_LOG=on and specify further paramters with the custom
```
miraclx, line 75:
```suggestion
/// NEAR_ENABLE_SANDBOX_LOG=1 and specify further paramters with the custom
```

workspaces/src/network/server.rs
```diff
@@ -67,3 +70,16 @@ impl Drop for SandboxServer {
...
+/// NEAR_SANDBOX_LOG for higher levels of specificity. NEAR_SANDBOX_LOG args
+/// will be forward into RUST_LOG environment variable as to not conflict
+/// with similar named log targets.
+fn supress_sandbox_logs_if_required() {
+    if let Err(_) = std::env::var("NEAR_ENABLE_SANDBOX_LOG") {
+        // non-exhaustive list of targets to supress, since choosing a default LogLevel
+        // does nothing in this case, since nearcore seems to be overriding it somehow:
+        std::env::set_var("NEAR_SANDBOX_LOG", "near=error,stats=error,network=error");
+    };
```
miraclx, line 84: With this, `NEAR_ENABLE_SANDBOX_LOG=0` would still activate sandbox logging.

PR #89, opened 2022-03-09
fix resolution of relative paths when compiling contracts

miraclx:
Implements resolution for the issue pointed out in https://github.com/near/workspaces-rs/pull/77#discussion_r820755762

> TBH, this path is a bit confusing.. it's not immediately clear that it's relative to the `workspaces` folder (it reads as though this path points to `workspaces/tests/tests/test-contracts/status-message`).
> 
> Ideally, given the location of this file, shouldn't the contract path be `./test-contracts/status-message`?
> 
> I'm thinking maybe we should wrap the `compile_project` function in a macro like `compile!()` that internally checks `file!()`, getting the parent and resolving relative paths from there (leaving paths as-is if they're already absolute).
> 
> Something like this:
> 
> ```rust
> #[macro_export]
> macro_rules! compile {
>     ($contract_path:expr) => {
>         $crate::__compile_contract(::std::path::Path::new(file!()), $contract_path)
>     };
> }
> ```
> 
> Although there's an issue when using a cargo workspace with members (as with the case of `workspaces-rs`), because `file!()` returns the path relative to the root workspace and not the member crate itself. So, from `deploy_project.rs`, `parent(file!())` returns `workspaces/tests`
> 
> I'm thinking we could use `cargo-metadata` to first, return all the members, with their respective paths, we can then find the member that owns the path returned by `parent(file!())` and resolve the relative path from there.
> 
> Although, you'd have to introduce new dependencies like [`pathdiff`](https://docs.rs/pathdiff/0.2.1/pathdiff/fn.diff_paths.html) to resolve relative paths, seeing as the stdlib doesn't have that functionality.

### DIFF
--- a/workspaces/src/cargo/mod.rs
+++ b/workspaces/src/cargo/mod.rs
@@ -142,13 +142,13 @@ impl<P: AsRef<Path>> __ContractCompiler<P> {
 ///
 /// - Relative paths:
 ///
-///   ```no_run
+///   ```ignore
 ///   let wasm = compile_contract!("../../contract");
 ///   ```
 ///
 /// - Absolute paths:
 ///
-///   ```no_run
+///   ```ignore
 ///   let wasm = compile_contract!("/workspace/contract");
 ///   ```
 fn resolve_path<P: AsRef<Path>>(caller: &Path, contract_path: P) -> anyhow::Result<PathBuf> {
--- a/workspaces/src/cargo/mod.rs
+++ b/workspaces/src/cargo/mod.rs
@@ -3,8 +3,7 @@ use async_process::Command;
 use cargo_metadata::{Message, Metadata, MetadataCommand};
 use std::env;
 use std::fmt::Debug;
-use std::fs;
-use std::path::Path;
+use std::path::{Component, Path, PathBuf};
 use std::process::Stdio;
 use tracing::debug;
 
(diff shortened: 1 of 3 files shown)


itegulov:
I haven't looked through the implementation yet, will try to do a pass tomorrow. For now I have just a general question about the reasoning here:

> TBH, this path is a bit confusing.. it's not immediately clear that it's relative to the workspaces folder (it reads as though this path points to workspaces/tests/tests/test-contracts/status-message).
> 
> Ideally, given the location of this file, shouldn't the contract path be ./test-contracts/status-message?

I am a bit confused here. Don't all runtime functions behave like that? For example `tokio::fs::read` accepts a path relative to the root project directory as well.

I have only encountered the behavior you suggest in macros (e.g. `include_bytes!`), but does it make sense for this feature to have the same behavior if all of the work happens at runtime?

miraclx:
> I have only encountered the behavior you suggest in macros (e.g. include_bytes!)

Hey, yeah, you're right, I guess I'm just leaving this here as an alternative as it feels more intuitive. We can have the discussion of whether we actually want this.

I also considered the implications at runtime vs. at compile time. But I figured since the target domain for `workspaces-rs` is in testing, then it's unlikely that people would;

1. move the binary around and execute it in a different working directory.
2. execute the binary manually.
    
    Seeing as `cargo test --test deploy_project` works, but `./target/debug/deps/deploy_project--` doesn't. (with or without this patch).
    
    It seems that `cargo`, by default, sets the current working directory to the crate root (not workspace root).
    So, to execute the binary manually and have it execute successfully, you have to `cd workspaces && ../target/debug/deps/deploy_project--`

And finally, there's a chance that using a `compile_contract!()` macro over the `compile_project()` function might give people the impression that their contracts are being compiled and loaded at compile time (à la `include_bytes!()`) instead of at runtime.

I'm not sure what's the better way out. But this PR at least helps the current state be a tad less confusing.

austinabell:
> > I have only encountered the behavior you suggest in macros (e.g. include_bytes!)
> 
> Hey, yeah, you're right, I guess I'm just leaving this here as an alternative as it feels more intuitive. We can have the discussion of whether we actually want this.
> 
> I also considered the implications at runtime vs. at compile time. But I figured since the target domain for `workspaces-rs` is in testing, then it's unlikely that people would;
> 
> 1. move the binary around and execute it in a different working directory.
> 2. execute the binary manually.
>    Seeing as `cargo test --test deploy_project` works, but `./target/debug/deps/deploy_project--` doesn't. (with or without this patch).
>    It seems that `cargo`, by default, sets the current working directory to the crate root (not workspace root).
>    So, to execute the binary manually and have it execute successfully, you have to `cd workspaces && ../target/debug/deps/deploy_project--`
> 
> And finally, there's a chance that using a `compile_contract!()` macro over the `compile_project()` function might give people the impression that their contracts are being compiled and loaded at compile time (à la `include_bytes!()`) instead of at runtime.
> 
> I'm not sure what's the better way out. But this PR at least helps the current state be a tad less confusing.

Yeah, `fs` utilities within Rust and other crates makes sense to use the current working directory, but in this case, it might not make much sense in this context where the compilation would depend on the working directory. While I think it might be the right decision, we might want to use a proc macro to be able to do more compile-time checks to make sure it's a valid location (perhaps doing a `cargo check` or something more light if available to avoid only failing at runtime). 

While having a macro option that will reference it directly and do compile-time checks, I don't think the basic Rust API we expose should do this and should take on the default `Path` behaviour because someone might want to compile something based on the working directory. I suppose it's opinionated whether we support a basic API, macro, or both, but I don't think that this declarative macro and pseudo-private types/methods path is the cleanest solution.

Issue #86, opened 2022-03-08
docs.rs build failing with current version


austinabell:
https://docs.rs/crate/workspaces/0.1.1/builds/496003

This broke with 0.1.1 release, and we should fix this before it gets to 0.2

Failing because it tries to download sandbox binary by default

austinabell:
https://docs.rs/crate/workspaces/0.2.0/builds/537593 failed again for 0.2

ChaoticTempest:
yeah it looks like docs rs actually has their own specific build that I noticed. It requires us to reference an environment variable. Not too bad of a fix, but this time, let's get a local build running to just make sure. We can make a patch release later since this doesn't affect anything API-wise

austinabell:
Having a working docs.rs page is very important for people using this. We should prioritize this

Also might be a good time to scan the docs to make sure key things are documented

ChaoticTempest:
Definitely. I'll try getting something up tomorrow. The docs should be mostly covered. All things exposed should have docs covered at least

itegulov:
Closed by https://github.com/near/workspaces-rs/pull/115

Issue #87, opened 2022-03-08
Link readme to workspaces crate


austinabell:
No readme shows up on crates.io https://crates.io/crates/workspaces/0.1.1

Can just soft link the readme, as this seems to be the standard used with Rust crates like [serde](https://github.com/serde-rs/serde/blob/master/serde/README.md)

itegulov:
Fixed by https://github.com/near/workspaces-rs/pull/90

PR #90, opened 2022-03-09
chore: fix doc builds and adds README

ChaoticTempest:
Resolves https://github.com/near/workspaces-rs/issues/86 and https://github.com/near/workspaces-rs/issues/87

Also adds a feature flag to disable installing sandbox if a user just wants to use testnet only and isn't on a compatible OS

### DIFF
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -43,3 +43,7 @@ libc = "0.2"
 borsh = "0.9"
 test-log = { version = "0.2.8", default-features = false, features = ["trace"] }
 tracing-subscriber = { version = "0.3.5", features = ["env-filter"] }
+
+[features]
+default = ["install"]
+install = []  # Install the sandbox binary during compile time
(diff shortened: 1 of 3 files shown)


itegulov (approved):

workspaces/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "workspaces"
 version = "0.1.1"
 edition = "2018"
 license = "MIT OR Apache-2.0"
+readme = "README.md"
```
austinabell, line 6: pretty sure this is the default, but I guess nbd to have

miraclx (approved):

austinabell (approved):

workspaces/README.md
```diff
@@ -0,0 +1 @@
+../README.md
```
austinabell, line 1: oh wait this isn't linked correctly it seems

austinabell (changes requested):

workspaces/README.md
ChaoticTempest: Oh good catch. The newest commit should fix this

austinabell (approved):

Issue #62, opened 2022-01-25
Why does dev-deploy on sandbox make a throw away account?


willemneal:
This makes sense on testnet, but locally it just seems to generate extra files.

itegulov:
Hmm, what would the alternative be though?

> locally it just seems to generate extra files.

That is true, but I think the intended way to use workspaces is to just add `.near-credentials` to .gitignore and forget that the folder exists. Not sure what is the problem with those extra files other than taking a few MBs of space on your PC.

austinabell:
Well, I think at least an improvement to this is to clean up the credentials after they are used, or just never persist them to a file if only used for the test execution.

ChaoticTempest:
outdated. Current version does nothing about credentials and requires users to specify `account.store_credentials`

PR #98, opened 2022-03-30
feat: Moved credentials storage to user side

ChaoticTempest:
Now no longer polluting working directory with credentials constantly being saved. And now allowing users to store it themselves if they wish to save them.

Also fixed a minor thing with the README title having code formatting.

### DIFF
--- a/workspaces/src/network/account.rs
+++ b/workspaces/src/network/account.rs
@@ -1,3 +1,5 @@
+use std::path::Path;
+
 use near_primitives::views::AccountView;
 
 use crate::types::{AccountId, Balance, InMemorySigner};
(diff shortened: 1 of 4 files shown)


itegulov (approved):
I guess this closes https://github.com/near/workspaces-rs/issues/62?

ChaoticTempest:
> I guess this closes #62?

yeah, I was trying to find the issue related to it. Thanks for pointing to it

workspaces/src/network/account.rs
```diff
@@ -128,6 +130,19 @@ impl Account {
     ) -> Transaction<'a> {
         Transaction::new(worker.client(), self.signer().clone(), contract_id.clone())
     }
+
+    /// Store the credentials of this account locally in the directory provided.
+    pub fn store_credentials(&self, save_dir: impl AsRef<Path>) -> anyhow::Result<()> {
```
austinabell, line 135: we may want this to be async to avoid breaking changes in future when the io is done async, but nbd to just have a breaking change later

austinabell (approved):
LGTM! Seems like visibility of this might be low, wonder how we can share this exists

PR #99, opened 2022-04-01
fix: make users wait for sandbox to completely startup

ChaoticTempest:
Queries like `view_latest_block` can fail if the users directly call into sandbox right after being spunned up. So, this fixes that and makes `sandbox()` calls `async` where we wait for the sandbox to fully spin up. To make this consistent, other calls likes `testnet()` and `mainnet()` are also `async`hronous.


workspaces/src/worker/mod.rs
```diff
@@ -20,41 +20,41 @@ where
...
 /// a [`Worker`] that can interact with it.
-pub fn mainnet() -> Worker<Mainnet> {
+pub async fn mainnet() -> Worker<Mainnet> {
     Worker::new(Mainnet::new())
 }
 
 /// Connect to the [mainnet archival](https://near-nodes.io/intro/node-types#archival-node)
 /// network, and grab a [`Worker`] that can interact with it.
-pub fn mainnet_archival() -> Worker<Mainnet> {
+pub async fn mainnet_archival() -> Worker<Mainnet> {
```
austinabell, line 47: are these all made async for a specific reason? Also, why not do a status call on other networks other than sandbox and return a result for the other networks? Feels weird to have this inconsistency

ChaoticTempest: yeah I was thinking about consistency, but the Result one was because the other networks didn't return result, but lets go with this:
> feels like you should also make them return results and check the RPC status if so.

normally, RPC for these should never be down, but then again that assumption would be wrong if it ever goes down. I guess by returning a Result, the experience would be better in that particular case

austinabell: Yeah, I think having the UX better in that failing case, or simply having some issue connecting with the RPC for any other reason is enough to keep this in line. Might be confusing for developers also to have different signatures for these.

Opinionated comment, so feel free to disregard, your call

workspaces/src/network/server.rs
```diff
@@ -39,6 +44,17 @@ impl SandboxServer {
     pub fn rpc_addr(&self) -> String {
         format!("http://localhost:{}", self.rpc_port)
     }
+
+    /// Wait until the RPC is available for sandbox. Will error out if this exceeds the amount of retries.
+    pub(crate) async fn wait_for_rpc(client: &Client) -> anyhow::Result<()> {
+        let retry_six_times = std::iter::repeat_with(|| Duration::from_millis(500)).take(6);
+        Retry::spawn(retry_six_times, || async { client.status().await })
```
itegulov, line 51: Will this silently eat `JsonRpcError`s? Maybe we should log them?

itegulov (approved):

workspaces/src/worker/mod.rs
ChaoticTempest: went with it since it makes the whole experience much better. Thanks for the suggestion!

austinabell (approved):

PR #101, opened 2022-04-04
fix: move TLA creator trait to dev network

austinabell:
Having this fail at runtime is probably not ideal.


ChaoticTempest (approved):
Thanks for cleaning this up!

PR #102, opened 2022-04-04
refactor!: Reorganize exports and modules exposed

austinabell:
I'll try to summarize all changes:

- Export network module publicly, which includes the specific network implementations (it is necessary for these to be exported since it's part of public API)
- Move result module to root level (non-breaking, was exported before)
- Move network agnostic types to the `types` directory (non-breaking)
- Remove `StatePatcher` and `AllowStatePatcher` traits. Ideally, this came in separate PR, but didn't want to export them just to remove later
- Create `operations` module. I couldn't think of a better name but these builder types need to be exported since part of the public API. Decided to include the breaking change of removing the root declaration of `Function`, but this can be re-exported if we think it should be root level

I don't think this is perfect, but moves towards exports being a little more organized for maintainability but also so that the types exposed on our public API are exported.


workspaces/src/network/mainnet.rs
```diff
@@ -3,10 +3,11 @@ use std::path::PathBuf;
 use async_trait::async_trait;
 
 use crate::network::Info;
-use crate::network::{Account, CallExecution, NetworkClient, NetworkInfo, TopLevelAccountCreator};
+use crate::network::{NetworkClient, NetworkInfo, TopLevelAccountCreator};
+use crate::result::CallExecution;
```
willemneal, line 7: Why is this type not exposed?  It is returned by exposed functions.  So if a user wants to make a function that returns the result of one of the exposed function they reference the type.

austinabell: What are you referring to? This is just a type that wraps the type returned that includes metadata

willemneal: Sorry I meant from the crate: https://github.com/near/workspaces-rs/pull/102/files#diff-f2509a15c16bf20cbd905fc2d19c6b77363687408bf1020327da68936ec19ef9

Account exposes `CallExecutionDetails` and `CallExecution`, but they aren't exported.

ChaoticTempest: those are exposed under the `result` module which we have exported via `pub mod result;` in `lib.rs`

willemneal: Ah I see thanks!

ChaoticTempest (approved):
new changes lgtm

PR #104, opened 2022-04-05
release: 0.2

ChaoticTempest:


workspaces/Cargo.toml
```diff
@@ -11,7 +11,7 @@ Library for automating workflows and testing NEAR smart contracts.
 
 [dependencies]
 async-trait = "0.1"
-async-process = { version = "1.3.0", optional = true }
+async-process = "1.3"
```
austinabell, line 14: Why is this not optional anymore? Doesn't seem like you've removed the unstable feature config from the only code that uses it?

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+- TLA trait no longer apart of all networks -- only dev-networks (sandbox, testnet). https://github.com/near/workspaces-rs/pull/101
+- Retry times have now been shorten and should take a maximum of 1 second. https://github.com/near/workspaces-rs/pull/92
+- doc builds on [docs.rs](https://docs.rs) has now been fixed. https://github.com/near/workspaces-rs/pull/90
+- Empty JSON array is no longer a valid default argument supplied to transactions. Recommended to supply empty `{}` in the case of JSON if all function arguments in the contract are optional types. https://github.com/near/workspaces-rs/pull/84
+- `patch_state` now takes in slices. https://github.com/near/workspaces-rs/pull/80 and https://github.com/near/workspaces-rs/pull/79
+- Make `access_key` call do optimistic queries which led to better retry times. https://github.com/near/workspaces-rs/pull/60
+- Functions no longer take in owned but referenced `AccountId`s now. https://github.com/near/workspaces-rs/pull/52
+
+## `0.1.1` [01-24-2021]
```
austinabell, line 31:
```suggestion
## [`0.2.0`] - 04-05-2022

### Features
- Time-traveling - the ability to go forwards & backwards in block height within tests. This allows to test time specific data changing within contracts: https://github.com/near/workspaces-rs/pull/73
- Credentials created from account/contract creation are now allowed to be stored and specified by users. https://github.com/near/workspaces-rs/pull/98
- [Unstable] Allow users to compile contract projects within tests without having to manually go through this step. https://github.com/near/workspaces-rs/pull/77
- Batch transactions or transactions with multiple actions are now possible. https://github.com/near/workspaces-rs/pull/72
- Sandbox node (nearcore binary) logs are now suppressed and can be re-enabled if desired. https://github.com/near/workspaces-rs/pull/85
- Results now expose logs, receipts, and transaction outcome values. https://github.com/near/workspaces-rs/pull/70
- Convenience methods `Worker::view_code`, `Worker::view_latest_block`, `Worker::view_account`, `Account::view_account`, `Contract::view_account`, `Contract::view_code` now available. https://github.com/near/workspaces-rs/pull/82
- Improve error handling. If a transaction fails, this error will now be apart of the `Result` return initially. https://github.com/near/workspaces-rs/pull/83
- Added `tracing` logging to internal code and examples. https://github.com/near/workspaces-rs/pull/55 and https://github.com/near/workspaces-rs/pull/75
- Convenient `CallExecutionDetails::{is_success, is_failure}` for testing outcomes of transactions. https://github.com/near/workspaces-rs/pull/58
- Added `mainnet_archival` and `testnet_archival`, where `ref-finance` example now uses `mainnet_archival`. https://github.com/near/workspaces-rs/pull/57 and https://github.com/near/workspaces-rs/pull/94


### Fixes
- No longer running into non-deterministic query failures if RPC isn't available, but this is a breaking API. All `workspaces::{sandbox, testnet, mainnet}` now require `.await?` at the end. https://github.com/near/workspaces-rs/pull/99
- TLA trait no longer apart of all networks -- only dev-networks (sandbox, testnet). https://github.com/near/workspaces-rs/pull/101
- Retry times have now been shorten and should take a maximum of 1 second. https://github.com/near/workspaces-rs/pull/92
- doc builds on [docs.rs](https://docs.rs) has now been fixed. https://github.com/near/workspaces-rs/pull/90
- Empty JSON array is no longer a valid default argument supplied to transactions. Recommended to supply empty `{}` in the case of JSON if all function arguments in the contract are optional types. https://github.com/near/workspaces-rs/pull/84
- `patch_state` now takes in slices. https://github.com/near/workspaces-rs/pull/80 and https://github.com/near/workspaces-rs/pull/79
- Make `access_key` call do optimistic queries which led to better retry times. https://github.com/near/workspaces-rs/pull/60
- Functions no longer take in owned but referenced `AccountId`s now. https://github.com/near/workspaces-rs/pull/52

## [`0.1.1`] - 01-24-2021
```

Format is wrong, realized I have it wrong in the SDK so not sure if you pulled from there.

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Fixes
+- Fix race condition when installing sandbox and running multiples tests at the same time. https://github.com/near/workspaces-rs/pull/46
+
+
+[Unreleased]: https://github.com/near/near-sdk-rs/compare/0.2.0...HEAD
+[0.2.0]: https://github.com/near/near-sdk-rs/compare/0.1.1...0.2.0
+[0.1.1]: https://github.com/near/near-sdk-rs/compare/0.1.0...0.1.1
+[0.1.0]: https://github.com/near/near-sdk-rs/releases/tag/0.1.0
```
austinabell, line 40: This is linking toward the SDK, check to make sure these links resolve

workspaces/Cargo.toml
itegulov: I am guessing because the new version of near-sandbox-utils pulls it in anyway

ChaoticTempest: oh wait, was originally gonna swap over to the async version, but then that was gonna be a huge process due to async drop and stuff. So this was remnant of that. It should be optional again since its no longer being used

austinabell: I don't think that's a good reason to import it again if not needed directly, is it?

ChaoticTempest: yeah sorry, I didn't mean to say that was the reason, but meant I'll remove it

itegulov: I think Austin was referring to my comment :)

austinabell: > I think Austin was referring to my comment :)

I was, for some reason my comments don't update in real-time :D 

not a big deal either way

CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+- Sandbox node (nearcore binary) logs are now suppressed and can be re-enabled if desired. https://github.com/near/workspaces-rs/pull/85
+- Results now expose logs, receipts, and transaction outcome values. https://github.com/near/workspaces-rs/pull/70
+- Convenience methods `Worker::view_code`, `Worker::view_latest_block`, `Worker::view_account`, `Account::view_account`, `Contract::view_account`, `Contract::view_code` now available. https://github.com/near/workspaces-rs/pull/82
+- Improve error handling. If a transaction fails, this error will now be apart of the `Result` return initially. https://github.com/near/workspaces-rs/pull/83
+- Added `tracing` logging to internal code and examples. https://github.com/near/workspaces-rs/pull/55 and https://github.com/near/workspaces-rs/pull/75
+- Convenient `CallExecutionDetails::{is_success, is_failure}` for testing outcomes of transactions. https://github.com/near/workspaces-rs/pull/58
+- Added `mainnet_archival` and `testnet_archival`, where `ref-finance` example now uses `mainnet_archival`. https://github.com/near/workspaces-rs/pull/57 and https://github.com/near/workspaces-rs/pull/94
+
+
```
austinabell, line 20: Might want to add a changes column that has #99, #101, and #102

Also should include #109 in fixes or in `Changes`

ChaoticTempest: I'll move those around

workspaces/Cargo.toml
```diff
@@ -51,4 +51,4 @@ tracing-subscriber = { version = "0.3.5", features = ["env-filter"] }
 [features]
 default = ["install"]
 install = []  # Install the sandbox binary during compile time
-unstable = ["cargo_metadata", "async-process"]
\ No newline at end of file
+unstable = ["cargo_metadata", "async_process"]
```
austinabell, line 54:
```suggestion
unstable = ["cargo_metadata", "async-process"]
```

ChaoticTempest: oops

README.md
```diff
@@ -13,50 +13,120 @@
...
+## Release notes
+**Release notes and unreleased changes can be found in the [CHANGELOG](CHANGELOG.md)**
+
 ## Requirements
 - rust v1.56 and up
-- MacOS (x86), M1 (through rosetta) or Linux (x86) for sandbox tests. Testnet is available regardless
+- MacOS (x86) or Linux (x86) for sandbox tests. Testnet is available regardless
+
+### M1 MacOS
+NOTE: Current version of workspaces does not provide support for M1 macbooks, due to some internal upgrades with wasmer. Look at using for workspaces version `0.1.1` for now until we get this settled out or check the progress of it with this [issue](https://github.com/near/workspaces-rs/issues/110).
```
AustinBaggio, line 24: NOTE: Current version of workspaces-rs does not support use on M1 chip devices due to internal upgrades with wasmer. M1 users should use workspaces-rs version `0.1.1` until this problem has been addressed. Check the progress with on this issue [here](https://github.com/near/workspaces-rs/issues/110).

README.md
```diff
@@ -13,50 +13,120 @@
...
 
-### M1 MacOS Setup
```
AustinBaggio, line 20: Should this section be deleted for now? Since it doesn't work?

README.md
```diff
@@ -13,50 +13,120 @@
...
-    assert_eq!(result, "OUR_EXPECTED_RESULT");
+    println!("--------------\n{}", result);
+    println!("Dev Account ID: {}", contract.id());
     Ok(())
 }
 ```
 
 ## Examples
-Some examples can be found in `examples/src/*.rs` to run it standalone.
+Some further examples can be found in `examples/src/*.rs` to be ran standalone.
```
AustinBaggio, line 127: More standalone examples can be found in `examples/src/*.rs`

README.md
```diff
@@ -13,50 +13,120 @@
   </p>
 </div>
 
+## Release notes
+**Release notes and unreleased changes can be found in the [CHANGELOG](CHANGELOG.md)**
+
 ## Requirements
 - rust v1.56 and up
```
AustinBaggio, line 20: - Rust v1.56 and up

README.md
ChaoticTempest: It works for `0.1.1`, so we should probably keep it for those users

README.md
ChaoticTempest: as you can see, I'm no English major haha

austinabell (approved):
ignoring wording nits on README, looks good to be released

README.md
```diff
@@ -13,50 +13,120 @@
...
 To be able to use this library on an M1 Mac, we would need to setup rosetta plus our cross compile target:
 ```
 softwareupdate --install-rosetta
 rustup default stable-x86_64-apple-darwin
 ```
 
-## Testing
-A simple test to get us going and familiar with the features:
+### Note about compiling with contract
+`workspaces-rs` does not currently support compiling to WASM. So, if we are compiling this library alongside a `wasm32` target such as compiling alongside a NEAR contract, it is best if we put this dependency in `[dev-dependencies]` section to avoid any conflicts.
```
doriancrutcher, line 33: Discussed in chat. But I initially read this as a reference to the contract code being complied rather than workspaces itself

README.md
```diff
@@ -13,50 +13,120 @@
...
+```rust
 
 #[tokio::test]
-async fn test_deploy_and_view() -> anyhow::Result<()> {
+async fn test_nft_contract() -> anyhow::Result<()> {
     let worker = workspaces::sandbox().await?;
+    let wasm = std::fs::read(NFT_WASM_FILEPATH)?;
+    let contract = worker.dev_deploy(&wasm).await?;
+```
+where
```
doriancrutcher, line 70: Small typo "Where"

README.md
ChaoticTempest: @doriancrutcher @AustinBaggio should look better now

doriancrutcher (commented):
Only added a note for clarification of one section and mentioned some typos but overall I think it looks good!

README.md
```diff
@@ -17,20 +17,35 @@
...
 ```
 
-### Note about compiling with contract
-`workspaces-rs` does not currently support compiling to WASM. So, if we are compiling this library alongside a `wasm32` target such as compiling alongside a NEAR contract, it is best if we put this dependency in `[dev-dependencies]` section to avoid any conflicts.
+### Note about compiling with NEAR Contracts
+`workspaces-rs`, the library itself, does not currently compile to WASM. So, if we were compiling it alongside a `wasm32` target, such as compiling alongside a NEAR contract:
+```toml
+# Cargo.toml
+[dependencies]
+workspaces = "*"
```
austinabell, line 37: Why are we showing an example of a broken pattern in the readme?

README.md
```diff
@@ -17,20 +17,35 @@
...
+near-sdk = "*"
+```
+It would throw an error when we run `cargo build --target wasm32-unknown-unknown`. Instead, we should add `workspaces-rs` into `[dev-dependencies]` for testing contracts:
+```toml
+# Cargo.toml
+[dependencies]
+near-sdk = "*"
+
+[dev-dependencies]
+workspaces = "*"
```
austinabell, line 47: The 0.2 version probably should have been included here, no reason to wildcard

README.md
ChaoticTempest: Just trying to show the dichotomy of how it'd look like if it didn't work vs the working example. Any suggestions to make it better?

README.md
ChaoticTempest: yeah, just didn't want to miss having to update the text here when we update/release past "0.2". I can change this to just `workspaces = ...` just so readers don't confuse it with valid syntax

README.md
austinabell: This just seems like something relating to the SDK. I feel like if we just share that it does not compile to wasm with some information about that would be sufficient, rather than showing an example of workspaces being included in the SDK dependencies (which it should never be).

Opinionated so nothing has to change, just sharing thoughts that it might be confusing to someone reading

ChaoticTempest: I see, yeah that's a good point. I still mentioned the SDK since that seems to be the most common case people would hit from what I've seen. Added the fix here: https://github.com/near/workspaces-rs/pull/115

PR #105, opened 2022-04-05
fix: avoid creating new client for every request

austinabell:
Not sure if there is some context I'm missing for why this isn't kept, but this seems to avoid a lot of unnecessary initialization/allocations

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -28,15 +28,17 @@ pub(crate) const DEFAULT_CALL_DEPOSIT: Balance = 0;
 const ERR_INVALID_VARIANT: &str =
     "Incorrect variant retrieved while querying: maybe a bug in RPC code?";
 
-/// A client that wraps around JsonRpcClient, and provides more capabilities such
+/// A client that wraps around [`JsonRpcClient`], and provides more capabilities such
 /// as retry w/ exponential backoff and utility functions for sending transactions.
 pub struct Client {
-    rpc_addr: String,
+    rpc_client: JsonRpcClient,
 }
 
 impl Client {
-    pub(crate) fn new(rpc_addr: String) -> Self {
-        Self { rpc_addr }
+    pub(crate) fn new(rpc_addr: &str) -> Self {
+        Self {
+            rpc_client: JsonRpcClient::connect(rpc_addr),
+        }
     }
 
     pub(crate) async fn query_broadcast_tx(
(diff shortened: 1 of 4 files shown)


workspaces/src/rpc/client.rs
```diff
@@ -47,7 +49,7 @@ impl Client {
         near_jsonrpc_primitives::types::transactions::RpcTransactionError,
     > {
         retry(|| async {
-            let result = JsonRpcClient::connect(&self.rpc_addr).call(method).await;
+            let result = self.rpc_client.call(method).await;
```
itegulov, line 52: I think I have played around with this in the past, but I was getting https://github.com/hyperium/hyper/issues/2112 at runtime. If examples and tests work for you with this change I might have been doing something wrong.

austinabell: yeah, everything works on my end, possibly it was an old version of jsonrpc-client? cc @miraclx 

I can dig more into this tomorrow to see what's happening internally, but if there is doubt, this shouldn't come in 0.2 likely

ChaoticTempest: similar reason why I kept it as `connect` since it was erroring out if I tried reusing the client back then

miraclx: Yeah, this should've worked right from time. The likely errors from early versions of the jsonrpc-client probably would've been from `JsonRpcClient::call` taking `self` instead of `&self` that would've allowed you to trivially reuse the client for multiple RPC method calls. Technically, you could've safely `clone`d the client back then since it used an `Arc` underneath, but it wasn't very clear from the type system.

Then again, multiple calls to `connect` is not so bad, since the internal `Client` is reused across multiple `connect` calls.

It's useful, however, to share a single `RpcClient` instance, which saves on extra heap allocations needed to store the server address and request headers.

miraclx (approved):

ChaoticTempest:
@miraclx was testing this recently, but it seems like I'm getting a lot of `dispatch dropped without returning error` errors while running RPC against testnet. According to https://github.com/hyperium/hyper/issues/2112, it's because of the global/default connector. And it indeed seems to be so, since moving from
```rust
let rpc_client = JsonRpcClient::connect(rpc_addr);
```
into
```rust
let rpc_client = connector.connect(rpc_addr);
```
alleviates and solves the dispatch error I'm seeing

ChaoticTempest:
closing this one for https://github.com/near/workspaces-rs/pull/154

PR #109, opened 2022-04-05
fix!: Key type of state map and remove exposed primitives type

austinabell:
- Removes `StoreKey` from public interface for using prefixes
- Changes state map to have keys as bytes (there is nothing enforcing these to be utf8, this is incorrect)
- Adds helper function to contract struct to retrieve state

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -167,19 +167,15 @@ impl Client {
     pub(crate) async fn view_state(
         &self,
         contract_id: AccountId,
-        prefix: Option<StoreKey>,
-    ) -> anyhow::Result<HashMap<String, Vec<u8>>> {
-        self.view_state_raw(contract_id, prefix, None)
-            .await?
-            .into_iter()
-            .map(|(k, v)| Ok((String::from_utf8(k)?, v.to_vec())))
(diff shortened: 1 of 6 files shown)


workspaces/src/rpc/client.rs
```diff
@@ -170,19 +170,15 @@ impl Client {
...
-    ) -> anyhow::Result<HashMap<String, Vec<u8>>> {
-        self.view_state_raw(contract_id, prefix, None)
-            .await?
-            .into_iter()
-            .map(|(k, v)| Ok((String::from_utf8(k)?, v.to_vec())))
-            .collect()
+        prefix: Option<&[u8]>,
+    ) -> anyhow::Result<HashMap<Vec<u8>, Vec<u8>>> {
+        self.view_state_raw(contract_id, prefix, None).await
     }
```
ChaoticTempest, line 176: we can probably just delete this now, since it's the same as `view_state_raw` and rename `view_state_raw` as `view_state`

austinabell: Sure! They are slightly different APIs so I kept, but I'll remove now since all internal

ChaoticTempest (approved):

PR #111, opened 2022-04-05
refactor!: Remove need to pass worker into every operation

### LINKED ISSUE #37: Do not require `&worker` when executing `{contract, account}.call`
`Account`s and `Contract`s do not need to take in worker when performing calls, but would end up requiring type parameters in the current state of the API. Once `Worker` no longer takes a type parameter and is resolved with this issue [https://github.com/near/workspaces-rs/issues/31](https://github.com/near/workspaces-rs/issues/31), then we can just add a `Worker` pretty simply into these types.

austinabell:
Alternative to #107, I personally think this approach is better, but the traits still need to be cleaned up as they are hard to use.

Closes #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ async fn test_deploy_and_view() -> anyhow::Result<()> {
         .await
         .expect("could not dev-deploy contract");
 
-    let result: String = contract.call(&worker, "function_name")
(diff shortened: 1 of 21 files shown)


austinabell:
Closing in favour of #181

PR #115, opened 2022-04-08
hotfix: doc builds on docs.rs

ChaoticTempest:
Been trying for over a day to get `docs.rs` builds to run locally, but keep running into issues with trying to reproduce their builds like unable to find dependencies in the docker container. So felt like it's better just to not waste anymore time trying to configure that and make a release to see if actually passes this time. `DOCS_RS` looks to be the way according to multiple sources so let's go with that. It doesn't affect users who are using the library anyways, but better to get this in sooner than later so people can view the docs.

Also, added some more docs that were missing from the module level

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,23 +29,8 @@ softwareupdate --install-rosetta
 rustup default stable-x86_64-apple-darwin
 ```
 
-### Note about compiling with NEAR Contracts
-`workspaces-rs`, the library itself, does not currently compile to WASM. So, if we were compiling it alongside a `wasm32` target, such as compiling alongside a NEAR contract:
-```toml
-# Cargo.toml
-[dependencies]
-workspaces = "*"
-near-sdk = "*"
-```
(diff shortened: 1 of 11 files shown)


austinabell (approved):
Wonder if we can release something with this pattern as a separate crate just to test to make sure it's valid before releasing this again

itegulov:
Yeah maybe let's try publishing this as another crate (`phuong-publish-test` or whatever) and see if it works?

austinabell:
> Yeah maybe let's try publishing this as another crate (`phuong-publish-test` or whatever) and see if it works?

I think there are some checks that duplicate code isn't pushed to crates.io, can't remember by what criteria, so might want to create a minimal crate with just this `build.rs` file and related logic if you want to be super safe

ChaoticTempest:
I'll give publish to a temp crate a go today. Also, been duo publishing to near-workspaces crate as well so I don't think duplicate code is problem haha

ChaoticTempest:
yup, that worked

Issue #114, opened 2022-04-08
Add betanet support


austinabell:
Motivation is related to https://github.com/near/nearcore/issues/6150#issuecomment-1092636701 but this would be generally useful for people to be able to test things with most cutting edge features

PR #116, opened 2022-04-12
feat: allow Worker<Betanet> creation

ChaoticTempest:
resolves https://github.com/near/workspaces-rs/issues/114.

This adds betanet support to workspaces, while the more general custom network creation is being addressed in a later PR

### DIFF
--- a/workspaces/src/worker/mod.rs
+++ b/workspaces/src/worker/mod.rs
@@ -2,7 +2,7 @@ mod impls;
 
 use std::sync::Arc;
 
-use crate::network::{Mainnet, Sandbox, Testnet};
+use crate::network::{Betanet, Mainnet, Sandbox, Testnet};
 use crate::Network;
 
 pub struct Worker<T> {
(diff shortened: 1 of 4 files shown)


workspaces/src/worker/mod.rs
```diff
@@ -53,6 +53,11 @@ pub async fn mainnet_archival() -> anyhow::Result<Worker<Mainnet>> {
     Ok(Worker::new(Mainnet::archival().await?))
 }
 
+/// Connect to the betanet network, and grab a [`Worker`] that can interact with it.
+pub async fn betanet() -> anyhow::Result<Worker<Betanet>> {
```
austinabell, line 57: note that you didn't add the `with_betanet` option to this, but probably fine without

austinabell (approved):

workspaces/src/worker/mod.rs
ChaoticTempest: added it since might as well consistent with it

ChaoticTempest: added it since might as well consistent with it

Issue #19, opened 2021-10-20
Support patching account and access-keys


ChaoticTempest:
Currently, only patching contract state is supported via https://github.com/near/workspaces-rs/pull/6. This is might also depend on https://github.com/near/workspaces-rs/issues/12 where patching contract state, accounts and access keys can have a common internal function.

PR #122, opened 2022-04-19
feat: patch account and access keys added

Buckram123:
Trying to implement #19. Not sure, how to make a friendly API for this. Thinking of using builder here.

### DIFF
--- a/workspaces/src/network/sandbox.rs
+++ b/workspaces/src/network/sandbox.rs
@@ -4,7 +4,11 @@ use std::str::FromStr;
 use async_trait::async_trait;
 use near_jsonrpc_client::methods::sandbox_fast_forward::RpcSandboxFastForwardRequest;
 use near_jsonrpc_client::methods::sandbox_patch_state::RpcSandboxPatchStateRequest;
+use near_primitives::account::AccessKey;
+use near_primitives::hash::CryptoHash;
 use near_primitives::state_record::StateRecord;
+use near_primitives::types::StorageUsage;
+use near_primitives::views::AccountView;
 
 use super::{AllowDevAccountCreation, NetworkClient, NetworkInfo, TopLevelAccountCreator};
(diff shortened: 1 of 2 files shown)


workspaces/src/worker/impls.rs
```diff
@@ -186,6 +187,21 @@ impl Worker<Sandbox> {
...
 
+    /// Patch account into the sandbox network, given an optional Account fields. This will allow
+    /// us to set amount or locked amount of NEAR tokens, change code hash, and storage usage.
+    pub async fn patch_account(
+        &self,
+        account_id: &AccountId,
+        amount: Option<Balance>,
+        locked: Option<Balance>,
+        code_hash: Option<CryptoHash>,
+        storage_usage: Option<StorageUsage>,
```
ChaoticTempest, line 198: > Not sure, how to make a friendly API for this. Thinking of using builder here.

yeah definitely, would be great if we can get a builder going for this. You can take a look at [patch.rs](https://github.com/near/workspaces-rs/blob/main/workspaces/src/rpc/patch.rs) to see how to get one going. Probably also best to put it there as well

workspaces/src/network/sandbox.rs
```diff
@@ -158,6 +162,70 @@ impl Sandbox {
         Ok(())
     }
 
+    pub(crate) async fn patch_account(
```
ChaoticTempest, line 165: Can you add tests for this and `patch_access_key`. Let me know if they aren't as straightforward to implement, but they should be simple if you follow the ones in [tests/patch_state.rs](https://github.com/near/workspaces-rs/blob/main/workspaces/tests/patch_state.rs).

ChaoticTempest (commented):
Thanks so much for implementing this! Just some minor things to add

Buckram123:
Closing in favor of #124.

PR #123, opened 2022-04-19
chore: Added cross contract tests

ChaoticTempest:
Added this specific test case where creating a top level account from a cross contract call should fail for a name that is too short in length


itegulov (approved):
That's an interesting behavior. Is there a way we can improve the error message from the SDK side?

austinabell:
Where is this behaviour documented? This seems very strange and want to understand it before we create a test case around it specifically

austinabell (approved):

ChaoticTempest:
@itegulov think it would require a protocol change since the host functions don't supply an error message when a Promise fails

Issue #12, opened 2021-10-13
Allow patch_state to take in multiple patches


ChaoticTempest:
Discussed in https://github.com/near/runner-rs/pull/6#discussion_r721570502. This would potentially require exposing some low level APIs for this to work due to not fulfilling the requirements of turning `BorshSerialize` into a trait objects for accepting multiple patches at the same time.

One potential solution mentioned by @austinabell is to have the `patch_state` signature take `IntoIterator<Item=(Key, Vec<u8>)>` then expose potentially higher level APIs with this.

Issue #88, opened 2022-03-08
Allow specifying sandbox version in workspaces


austinabell:
https://github.com/near/sandbox/issues/26 was added to sandbox to be able to configure the binary download location version, but it would be beneficial to be able to express this with some API from workspaces to override this without relying on setting an env variable when running tests.

I'll avoid suggesting a specific API since it depends on what happens with the sandbox issue and other issues/changes like https://github.com/near/workspaces-rs/pull/71

This issue is blocked until the direction of the sandbox issue is decided and implemented

Benefit for things like #81 to avoid having to use a specific version of workspaces that matches the sandbox version you want or managing downloading the binary manually

ChaoticTempest:
Would also like to note if we go in the direction of allowing users to specify versions, we'd most likely have to manage multiple sandbox versions at a time, and will likely need to adjust the sandbox crate accordingly to support that. And with https://github.com/near/workspaces-rs/pull/71 we'd want some consistency with the API to spin up sandbox the default way, the configured way, and the single instance way (if it makes it in)

austinabell:
Also have some thoughts on the workspaces and/or sandbox API

Can have a type like:

```rust
pub struct SandboxHandle(PathBuf);
```

and have `install` and `install_with_version` on sandbox return that type or just wrap it within workspaces, and this type could have the functions `init` and `run` and whatever else attached to it so you can be sure it's a valid binary you're using, instead of just expecting or assuming a download to the path will be done to use the binary?

Sorry if a bit scattered thoughts, can do a small PoC later

PR #124, opened 2022-04-21
feat: allow patch state to take in multiple patches + access key and account patching

adsick:
Hi, I'm working on #12 and #19. Now I feel like some review of what I've done would be great, please tell me what API is preferable and how to shape it (where to put stuff etc.).

Highlights:
* Improve over #80 
now accepting wider variety of types by using `Into<Vec<u8>>`
```rust
    pub async fn patch_state(
        &self,
        contract_id: &AccountId,
        key: impl Into<Vec<u8>>,
        value: impl Into<Vec<u8>>,
    )
```
* Implemented #12 
```rust
    pub async fn patch_state_multiple(
        &self,
        account_id: &AccountId,
        kvs: impl IntoIterator<Item = (impl Into<Vec<u8>>, impl Into<Vec<u8>>)>,
    )
```
* Implemented #19  by adapting #122 to be a builder
(for more see line 235 of sandbox.rs)
```rust
    /// Patch account state using builder pattern
    pub fn patch_account(&self, account_id: &AccountId) -> SandboxPatchStateAccountBuilder {
        self.workspace.patch_account(account_id.clone())
    }
```

Now I'm looking for some feedback to refine API :slightly_smiling_face:

### DIFF
--- a/workspaces/build.rs
+++ b/workspaces/build.rs
@@ -1,6 +1,9 @@
 fn main() {
     let doc_build = cfg!(doc) || std::env::var("DOCS_RS").is_ok();
     if !doc_build && cfg!(feature = "install") {
-        near_sandbox_utils::install().expect("Could not install sandbox");
+        // using unwrap because all the useful error messages are hidden inside
+        near_sandbox_utils::ensure_sandbox_bin().unwrap();
+        // previously the next line was causing near sandbox to be installed every time cargo build/check was called
+        // near_sandbox_utils::install().expect("Could not install sandbox");
     }
 }
--- a/workspaces/src/network/mod.rs
+++ b/workspaces/src/network/mod.rs
@@ -10,8 +10,9 @@ mod server;
 mod testnet;
 pub(crate) mod variants;
 
+pub(crate) use sandbox::SandboxPatchStateAccountBuilder; //not needed directly outside of the crate
+pub(crate) use sandbox::SandboxPatchStateBuilder; //not needed directly outside of the crate
 pub(crate) use variants::DEV_ACCOUNT_SEED;
-pub(crate) use sandbox::SandboxPatchStateBuilder; //not needed directly outside of the crate 
 
 pub use betanet::Betanet;
 pub use info::Info;
--- a/workspaces/src/network/mod.rs
+++ b/workspaces/src/network/mod.rs
@@ -11,12 +11,13 @@ mod testnet;
 pub(crate) mod variants;
 
 pub(crate) use variants::DEV_ACCOUNT_SEED;
+pub(crate) use sandbox::SandboxPatchStateBuilder; //not needed directly outside of the crate 
 
-pub use self::betanet::Betanet;
-pub use self::info::Info;
-pub use self::mainnet::Mainnet;
-pub use self::sandbox::Sandbox;
-pub use self::testnet::Testnet;
-pub use self::variants::{
+pub use betanet::Betanet;
+pub use info::Info;
+pub use mainnet::Mainnet;
+pub use sandbox::Sandbox;
+pub use testnet::Testnet;
+pub use variants::{
     AllowDevAccountCreation, DevAccountDeployer, NetworkClient, NetworkInfo, TopLevelAccountCreator,
 };
--- a/workspaces/src/network/sandbox.rs
+++ b/workspaces/src/network/sandbox.rs
@@ -5,7 +5,10 @@ use async_trait::async_trait;
 use near_jsonrpc_client::methods::sandbox_fast_forward::RpcSandboxFastForwardRequest;
 use near_jsonrpc_client::methods::sandbox_patch_state::RpcSandboxPatchStateRequest;
 use near_primitives::account::AccessKey;
+use near_primitives::hash::CryptoHash;
 use near_primitives::state_record::StateRecord;
+use near_primitives::types::StorageUsage;
+use near_primitives::views::AccountView;
 use std::iter::IntoIterator;
 
 use super::{AllowDevAccountCreation, NetworkClient, NetworkInfo, TopLevelAccountCreator};
@@ -14,7 +17,7 @@ use crate::network::Info;
 use crate::result::CallExecution;
 use crate::rpc::client::Client;
 use crate::rpc::patch::ImportContractTransaction;
-use crate::types::{AccountId, Balance, InMemorySigner, SecretKey, PublicKey};
+use crate::types::{AccountId, Balance, InMemorySigner, PublicKey, SecretKey};
 use crate::{Account, Contract, Network, Worker};
 
 // Constant taken from nearcore crate to avoid dependency
@@ -141,6 +144,12 @@ impl Sandbox {
         SandboxPatchStateBuilder::new(self, account_id)
     }
 
+    pub(crate) fn patch_account(&self, account_id: AccountId) -> SandboxPatchStateAccountBuilder {
+        SandboxPatchStateAccountBuilder::new(self, account_id)
+    }
+
+    // shall we expose convenience patch methods here for consistent API?
+
     pub(crate) async fn fast_forward(&self, delta_height: u64) -> anyhow::Result<()> {
         // NOTE: RpcSandboxFastForwardResponse is an empty struct with no fields, so don't do anything with it:
         self.client()
@@ -167,16 +176,12 @@ impl<'s> SandboxPatchStateBuilder<'s> {
     pub fn new(sandbox: &'s Sandbox, account_id: AccountId) -> Self {
         SandboxPatchStateBuilder {
             sandbox,
-            account_id: account_id,
+            account_id,
             records: Vec::with_capacity(4),
         }
     }
 
-    pub fn data(
-        mut self,
-        key: impl Into<Vec<u8>>,
-        value: impl Into<Vec<u8>>,
-    ) -> Self {
+    pub fn data(mut self, key: impl Into<Vec<u8>>, value: impl Into<Vec<u8>>) -> Self {
         let data = StateRecord::Data {
             account_id: self.account_id.clone(),
             data_key: key.into(),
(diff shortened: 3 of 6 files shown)


workspaces/src/network/sandbox.rs
```diff
@@ -135,36 +140,174 @@ impl Sandbox {
...
+            .await
+            .map_err(|err| anyhow::anyhow!("Failed to fast forward: {:?}", err))?;
+
+        Ok(())
+    }
+}
+
+//todo: review naming
+#[must_use = "don't forget to .apply() this `SandboxPatchStateBuilder`"]
+pub struct SandboxPatchStateBuilder<'s> {
```
Buckram123, line 167: I think the word `Sandbox` is redundant here. Probably best to put patch related builders in [patch.rs](https://github.com/near/workspaces-rs/blob/main/workspaces/src/rpc/patch.rs) as mentioned in [122](https://github.com/near/workspaces-rs/pull/122#discussion_r853591005)

workspaces/src/worker/impls.rs
```diff
@@ -174,16 +177,58 @@ impl Worker<Sandbox> {
...
     /// Patch state into the sandbox network, given a key and value. This will allow us to set
     /// state that we have acquired in some manner. This allows us to test random cases that
     /// are hard to come up naturally as state evolves.
     pub async fn patch_state(
         &self,
         contract_id: &AccountId,
-        key: &[u8],
-        value: &[u8],
+        key: impl Into<Vec<u8>>,
+        value: impl Into<Vec<u8>>,
```
ChaoticTempest, line 199: not everything will impl an `Into<Vec<u8>>` and by that notion won’t fit all cases. Best to keep as-is to avoid that situation, since in rust a lot of things can be converted into a series of bytes `&[u8]` including `Vec<u8>`. This will also avoid having to make a breaking change for this since now people would have to require this trait impl.

workspaces/src/worker/impls.rs
```diff
@@ -174,16 +177,58 @@ impl Worker<Sandbox> {
         self.workspace.import_contract(id, worker)
     }
 
+    /// Patch state into the sandbox network, using builder pattern. This will allow us to set
+    /// state that we have acquired in some manner. This allows us to test random cases that
+    /// are hard to come up naturally as state evolves.
+    pub fn patch_state_builder(&self, account_id: &AccountId) -> SandboxPatchStateBuilder {
```
ChaoticTempest, line 183: Think it's better just to have one `patch_state` function that way we can be consistent and have a defacto way of doing patching state. Let's merge this one and the other and just have this *builder* like one be the default

ChaoticTempest (changes requested):
Thanks so much for getting this implementation in! Did a cursory look and left a few comments so we can improve the maintainability of it

workspaces/build.rs
```diff
@@ -1,6 +1,9 @@
 fn main() {
     let doc_build = cfg!(doc) || std::env::var("DOCS_RS").is_ok();
     if !doc_build && cfg!(feature = "install") {
-        near_sandbox_utils::install().expect("Could not install sandbox");
+        // using unwrap because all the useful error messages are hidden inside
+        near_sandbox_utils::ensure_sandbox_bin().unwrap();
+        // previously the next line was causing near sandbox to be installed every time cargo build/check was called
+        // near_sandbox_utils::install().expect("Could not install sandbox");
```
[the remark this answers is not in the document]
adsick: Yes, I agree about the context part, but there also should be a note of 'precise' reason of failure.
I'll try to do something like "Could not install sandbox. Reason: ... "
That sounds good?

workspaces/src/network/sandbox.rs
```diff
@@ -135,36 +139,261 @@ impl Sandbox {
...
-    pub(crate) async fn patch_state(
+    pub(crate) fn patch_state(&self, account_id: AccountId) -> SandboxPatchStateBuilder {
+        SandboxPatchStateBuilder::new(self, account_id)
+    }
+
+    pub(crate) fn patch_account(&self, account_id: AccountId) -> SandboxPatchStateAccountBuilder {
+        SandboxPatchStateAccountBuilder::new(self, account_id)
+    }
+
+    pub(crate) fn patch_access_key(
```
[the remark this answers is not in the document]
adsick: Yeah, we didn't tested because we waited for the feedback. Tests will be added too.

workspaces/src/worker/impls.rs
adsick: > Let's merge this one and the other

which 'other' one?

workspaces/src/worker/impls.rs
austinabell: Yeah these traits are not as intuitive for a developer to read as well as it sticks us with using this API even when we remove the requirement for bytes to be converted to `Vec<u8>`

workspaces/src/worker/impls.rs
```diff
@@ -174,16 +177,58 @@ impl Worker<Sandbox> {
...
+    }
+
+    //todo: add more patch state methods like the previous one
+
+    /// Patch state into the sandbox network. Same as `patch_state` but accepts a sequence of key value pairs
+    pub async fn patch_state_multiple(
+        &self,
+        account_id: &AccountId,
+        kvs: impl IntoIterator<Item = (impl Into<Vec<u8>>, impl Into<Vec<u8>>)>,
     ) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
austinabell: yeah, for the same rationale as https://github.com/near/workspaces-rs/pull/124#discussion_r862990160 but primarily that the API will transition away from requiring owned `Vec<u8>` and will be a breaking change down the road to change. Prob best to constrain to `&[u8]` for now

workspaces/src/network/mod.rs
```diff
@@ -10,13 +10,16 @@ mod server;
...
-pub use self::mainnet::Mainnet;
-pub use self::sandbox::Sandbox;
-pub use self::testnet::Testnet;
-pub use self::variants::{
+pub use betanet::Betanet;
+pub use info::Info;
+pub use mainnet::Mainnet;
+pub use sandbox::Sandbox;
+pub use testnet::Testnet;
+pub use variants::{
```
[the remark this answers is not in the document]
austinabell: I'm a bit confused by this suggestion, what would cause this to break? This is just being more explicit about where it's coming from, if anything the latter would be open to being broken?

workspaces/src/network/sandbox.rs
```diff
@@ -135,36 +139,261 @@ impl Sandbox {
...
+impl<'s> SandboxPatchStateBuilder<'s> {
+    pub fn new(sandbox: &'s Sandbox, account_id: AccountId) -> Self {
+        SandboxPatchStateBuilder {
+            sandbox,
+            account_id,
+            records: Vec::with_capacity(4),
+        }
+    }
+
+    pub fn data(mut self, key: impl Into<Vec<u8>>, value: impl Into<Vec<u8>>) -> Self {
```
matklad, line 189: `impl Into<Vec<u8>>` looks weird. Is there some precedent for such API shape? I'd just use `Vec<u8>` or `&[u8]`, with a strong preference for `&[u8]`. 

`Vec<u8>` leaks impl details and forces needless allocation -- ideally, we don't store intermediate vecs at all and just build request string.

workspaces/src/network/sandbox.rs
```diff
@@ -135,36 +139,261 @@ impl Sandbox {
...
+            ref account_id,
+            ..
+        } = self;
+        records.extend(kvs.into_iter().map(|(key, value)| StateRecord::Data {
+            account_id: account_id.clone(),
+            data_key: key.into(),
+            value: value.into(),
+        }));
+        self
+    }
```
matklad, line 215: I'd rather let the caller to call `.data` in a loop. If we need multi-item API, I'd go for

```rust
impl std::iter::Extend<(&'_ [u8], &'_ [u8])> for SandboxPatchStateBuilder<'_> {

}
```

workspaces/src/network/sandbox.rs
matklad, line 189: Seems like we should have a way to *delete* key as well?

workspaces/src/network/sandbox.rs
```diff
@@ -135,36 +139,261 @@ impl Sandbox {
...
+        let Self {
+            ref mut records,
+            ref account_id,
+            ..
+        } = self;
+        records.extend(kvs.into_iter().map(|(key, value)| StateRecord::Data {
+            account_id: account_id.clone(),
+            data_key: key.into(),
+            value: value.into(),
+        }));
```
matklad, line 213: Minor: there's logical duplication between the two methods, one could delegate to the other (or both could delegate to internal _state_record)

workspaces/src/worker/impls.rs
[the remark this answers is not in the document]
matklad: You want `impl std::iter::Extend` here.

adsick:
Thanks a bunch for the feedback, @matklad 's guidance is looking great. We'll try to update this PR accordingly and then start testing when that's ok in terms of API and stuff.

workspaces/src/network/sandbox.rs
matklad: good point about the builder!

Indeed, to look at the std, theer's 

* https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.arg
* https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.args

Though, in this case, `data_multiple` looks like a particualr awkward name (data itself is plural).

Maybe we want `.entry` and `.entries` here? 

I'll delegate to you/dev-platform folks to make a judgement call here :)

workspaces/build.rs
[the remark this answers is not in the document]
ChaoticTempest: So after https://github.com/near/workspaces-rs/pull/134 we won't be able to change it to `ensure_sandbox_bin()`, but to alleviate this issue with re-installing, we'll also resolve it along with https://github.com/near/workspaces-rs/issues/88. It just requires a more involved versioning mechanism. Sounds good to you?

workspaces/src/network/sandbox.rs
ChaoticTempest: +1 for `.entry` and `.entries`

workspaces/src/worker/impls.rs
ChaoticTempest: I meant we can merge both patch_state and patch_state_builder to be just patch_state(AccountId) -> Builder. Would be a breaking API change, but it's fine since it would align with the other builder functions, and better to be consistent that way

workspaces/src/network/sandbox.rs
adsick: > +1 for `.entry` and `.entries`

got it

PR #126, opened 2022-04-21
example: croncat w/ fast forwarding

ChaoticTempest:
@TrevorJTClarke would greatly appreciate feedback from you for this example


examples/src/croncat.rs
```diff
@@ -0,0 +1,194 @@
...
+    println!("Registered agent details: {:#?}", registered_agent);
+    assert_eq!(registered_agent.status, AgentStatus::Active);
+    assert_eq!(&registered_agent.payable_account_id, agent.id());
+
+    // Advance 4500 blocks in the chain. 1 block takes approx 1.5 seconds to be produced, but we
+    // don't actually wait that long since we are time travelling to the future via `fast_forward`!
+    // After this `fast_forward` call, we should be ahead by about an hour, and it is expected for
+    // our agents to be able to execute the task we scheduled.
+    println!("Waiting until next slot occurs...");
+    worker.fast_forward(4500).await?;
```
TrevorJTClarke, line 126: Looks awesome! Would love to see it maybe do more forwarding - we have tasks that happen every hour for example, forward a few times proves out the agent model

TrevorJTClarke (commented):
Looks epic @ChaoticTempest! You def understand the full lifecycle of croncat!

examples/src/croncat.rs
ChaoticTempest: done!

examples/src/croncat.rs
```diff
@@ -0,0 +1,204 @@
+// This example will go over the croncat contracts found
+// [here](https://github.com/CronCats/contracts/blob/cafd3caafb91b45abb6e811ce0fa2819980d6f96/manager/src/lib.rs)
+// This will demonstrate a more involved example for fast-forwarding the blockchain to a future
+// state/time. This is useful for testing anything that is time dependent such as scheduling
+// This is perfect to showcase cron.cat which will schedule calling into contract functions
+// at a set amount of time we supply.
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use near_sdk::json_types::U128;
```
austinabell, line 9: prob shouldn't import and compile the SDK for just this and serde, wonder if there is a better alternative for this. Guess it doesn't have to come with this, but avoiding linking to SDK for examples is probably best

examples/src/croncat.rs
```diff
@@ -0,0 +1,204 @@
...
+use workspaces::{Account, AccountId, Contract, Worker};
+
+const MANAGER_CONTRACT: &[u8] = include_bytes!("../res/manager.wasm");
+const COUNTER_CONTRACT: &[u8] = include_bytes!("../res/counter.wasm");
+
+/// `AgentStatus` struct taken from [croncat repo](github.com/CronCats/contracts/) to
+/// deserialize into after we get the result of a transaction and converting over to
+/// this particular type.
+#[derive(BorshDeserialize, BorshSerialize, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(crate = "near_sdk::serde")]
```
austinabell, line 25: You don't need this if importing serde

examples/src/croncat.rs
```diff
@@ -0,0 +1,204 @@
...
+pub struct Agent {
+    pub status: AgentStatus,
+    pub payable_account_id: AccountId,
+    pub balance: U128,
+    pub total_tasks_executed: U128,
+    pub last_missed_slot: u128,
+}
+
```
austinabell, line 46: prob shouldn't import and compile the SDK for just this, prob better to just include the util here. Guess it doesn't have to come with this, but avoiding linking to SDK for examples is probably best.

Something like:

```suggestion
    #[serde(with = "string")]
    pub balance: u128,
    #[serde(with = "string")]
    pub total_tasks_executed: u128,
    pub last_missed_slot: u128,
}

mod string {
    use std::fmt::Display;
    use std::str::FromStr;

    use serde::{de, Serializer, Deserialize, Deserializer};

    pub fn serialize<T, S>(value: &T, serializer: S) -> Result<S::Ok, S::Error>
        where T: Display,
              S: Serializer
    {
        serializer.collect_str(value)
    }

    pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
        where T: FromStr,
              T::Err: Display,
              D: Deserializer<'de>
    {
        String::deserialize(deserializer)?.parse().map_err(de::Error::custom)
    }
}
```

Also, a bit unclear why the last_missed_slot would be an integer when the others aren't. I'm sure there is a motivation, but it isn't clear to me

austinabell (approved):
lgtm, added some comments, but nothing is hard blocking this

austinabell:
Also, prob not ideal to include such large wasm blobs, but I think it's fine for now

examples/src/croncat.rs
ChaoticTempest: Probably then we should provide our own set of these `U128`/`U64` types so people don't have to import SDK directly to make use of them. I'll put this in a separate PR

examples/src/croncat.rs
ChaoticTempest: > Also, a bit unclear why the last_missed_slot would be an integer when the others aren't. I'm sure there is a motivation, but it isn't clear to me

no idea, copied it directly from croncat. @TrevorJTClarke can probably explain here

examples/src/croncat.rs
austinabell: Those types are not the best to interact with. It would be best if there was a convenience module you can use in an attribute rather than using wrapper types.

workspaces/src/lib.rs
```diff
@@ -16,6 +16,7 @@ pub mod network;
 pub mod operations;
 pub mod prelude;
 pub mod result;
+pub mod serde;
```
austinabell, line 19: This module is going to collide with serde

ChaoticTempest: right, how about `deser` then, and then we can have `deser::serde_str`?

austinabell: I mean, this type of utility doesn't seem to be that necessary to be duplicated in workspaces. It's basically https://docs.rs/serde_with/1.13.0/serde_with/rust/display_fromstr/index.html and it might just be best for a lib like this to not be opinionated

ChaoticTempest: ahh didn't know that existed. I'll just replace that then and remove this

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [unreleased]
 
+### Added
+- Added additional example called `croncat.rs` to showcase more in depth `fast_forward` feature: https://github.com/near/workspaces-rs/pull/126
```
austinabell, line 6: note this does not change the API of the crate so this might not be something you want to include in the changelog

austinabell (approved):
new changes lgtm

CHANGELOG.md
ChaoticTempest: huh, so normally we don't want examples inside changelog? It make sense for internals being reworked around to not make it in the changelog, but I thought since example was public, it would be worth noting

austinabell: I mean, the point of the changelog is to see what has changed when deciding to update or see how the library was changed, but you can create whatever semantics if you think it would be beneficial. I don't think there is a rule against something like this, but I just can't intuitively see the benefit here.

cc @miraclx @itegulov do you have any thoughts about if example changes should be included in our changelogs to keep things consistent?

itegulov: I agree with Austin here. Don't think examples should be included in the changelog, I view them as a separate entity from the library itself.

ChaoticTempest: hmm I see, I'll just revert then

Issue #127, opened 2022-04-28
How to get block height?


ChzenChzen:
Is it possible to get block height for an executed call? I see only block hash.

ChaoticTempest:
This info is not directly available from the call. You'd have to do a separate call into getting a specific block via the block hash, but it seems `worker.view_block` isn't available. I'll add that now

ChzenChzen:
yeah, I've solved it:
```rust
async fn get_block_height(block_hash: CryptoHash) -> Result<u64> {
    let testnet_client = JsonRpcClient::connect(NEAR_TESTNET_ARCHIVAL_RPC_URL);
    let request = methods::block::RpcBlockRequest {
        block_reference: BlockReference::BlockId(BlockId::Hash(block_hash)),
    };
    let result = testnet_client.call(request).await?;
    Ok(result.header.height)
}
```
but it's better to have such option inside `workspaces-rs`

PR #129, opened 2022-04-29
feat: add view block

ChaoticTempest:
resolves https://github.com/near/workspaces-rs/issues/127 without forcing users to use `near-jsonrpc-client` to get block info

### DIFF
--- a/workspaces/src/types/mod.rs
+++ b/workspaces/src/types/mod.rs
@@ -214,3 +214,102 @@ impl From<AccessKey> for near_primitives::account::AccessKey {
         }
     }
 }
+
+/// Type representing the `block_id` paramter of an RPC query  that can be supplied into
+/// viewing blocks or chunks from the network.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum BlockId {
+    Height(BlockHeight),
+    Hash(CryptoHash),
+}
+
+impl From<BlockId> for near_primitives::types::BlockId {
+    fn from(id: BlockId) -> Self {
+        match id {
+            BlockId::Height(height) => Self::Height(height),
+            BlockId::Hash(hash) => Self::Hash(near_primitives::hash::CryptoHash(hash.0)),
+        }
+    }
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum Finality {
+    /// Optimistic finality
+    None,
+    /// Near-final finality
+    DoomSlug,
+    /// Final finality
+    Final,
+}
+
+impl From<Finality> for near_primitives::types::Finality {
+    fn from(finality: Finality) -> Self {
+        match finality {
+            Finality::None => Self::None,
+            Finality::DoomSlug => Self::DoomSlug,
+            Finality::Final => Self::Final,
+        }
+    }
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum SyncCheckpoint {
+    Genesis,
+    EarliestAvailable,
+}
+
+impl From<SyncCheckpoint> for near_primitives::types::SyncCheckpoint {
+    fn from(checkpoint: SyncCheckpoint) -> Self {
+        match checkpoint {
+            SyncCheckpoint::Genesis => Self::Genesis,
+            SyncCheckpoint::EarliestAvailable => Self::EarliestAvailable,
+        }
+    }
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum BlockReference {
+    BlockId(BlockId),
+    Finality(Finality),
+    SyncCheckpoint(SyncCheckpoint),
+}
+
+impl BlockReference {
+    /// The latest block produced in the blockchain, but not necessarily the latest
+    /// finalized block.
+    pub fn latest() -> Self {
+        Self::Finality(Finality::None)
+    }
+}
+
+impl From<BlockId> for BlockReference {
+    fn from(block_id: BlockId) -> Self {
+        Self::BlockId(block_id)
+    }
+}
+
+impl From<Finality> for BlockReference {
+    fn from(finality: Finality) -> Self {
+        Self::Finality(finality)
+    }
+}
+
+impl From<SyncCheckpoint> for BlockReference {
+    fn from(checkpoint: SyncCheckpoint) -> Self {
+        Self::SyncCheckpoint(checkpoint)
+    }
+}
+
+impl From<BlockReference> for near_primitives::types::BlockReference {
+    fn from(block_ref: BlockReference) -> Self {
+        match block_ref {
+            BlockReference::BlockId(id) => Self::BlockId(id.into()),
+            BlockReference::Finality(finality) => Self::Finality(finality.into()),
+            BlockReference::SyncCheckpoint(checkpoint) => Self::SyncCheckpoint(checkpoint.into()),
+        }
+    }
+}
--- a/workspaces/src/types/mod.rs
+++ b/workspaces/src/types/mod.rs
@@ -215,7 +215,7 @@ impl From<AccessKey> for near_primitives::account::AccessKey {
     }
 }
 
-/// Type representing the `block_id` paramter of an RPC query  that can be supplied into
+/// Type representing the `block_id` parameter of an RPC query that can be supplied into
 /// viewing blocks or chunks from the network.
 #[derive(Debug)]
 #[non_exhaustive]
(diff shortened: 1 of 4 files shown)


workspaces/src/worker/impls.rs
```diff
@@ -116,7 +117,15 @@ where
 
     /// View the latest block from the network
     pub async fn view_latest_block(&self) -> anyhow::Result<Block> {
-        self.client().view_block(None).await.map(Into::into)
+        self.view_block(BlockReference::latest()).await
+    }
+
+    /// View the block from the network given the block reference to it.
+    pub async fn view_block(&self, block_ref: BlockReference) -> anyhow::Result<Block> {
```
austinabell, line 124: This seems unergonomic to use, wonder if it's a better solution to split this function into this and `view_block_at_height`, `view_block_at_hash` or something so that someone doesn't have to write:

```rust
BlockReference::BlockId(BlockId::Height(_))
```

as a parameter and expose all of these enum types?

ChaoticTempest: they can directly do:
```
view_block(BlockId::Height(...).into())
```
to avoid importing `BlockReference` but that might be a little clunky. But BlockId should be the most frequently used case, so adding `view_block_at_height` and `view_block_at_hash` is not such a bad idea

austinabell: Yeah, I'm not sure what the best pattern would be for these, but maybe we should come up with a plan and stay consistent throughout the repo since this pattern will be common

itegulov (approved):

workspaces/src/worker/impls.rs
ChaoticTempest: I think what we have right now is pretty consistent where we have `BlockReference` being the general input, where we have helper methods for height and hash for the more used paths. I just gotta add the general method to importing contracts to make it consistent, but I'll push that to a separate PR since it's a big change in of itself

workspaces/src/worker/impls.rs
```diff
@@ -116,7 +117,25 @@ where
...
+        self.view_block(BlockReference::latest()).await
+    }
+
+    /// View the block at the supplied height
+    pub async fn view_block_at_height(&self, height: BlockHeight) -> anyhow::Result<Block> {
+        self.view_block(BlockId::Height(height).into()).await
+    }
+
+    /// View the block at the supplied hash
+    pub async fn view_block_at_hash(&self, hash: CryptoHash) -> anyhow::Result<Block> {
```
austinabell, line 129:
```suggestion
    pub async fn view_block_by_hash(&self, hash: CryptoHash) -> anyhow::Result<Block> {
```

Optional: this seems like a more logical name

workspaces/src/worker/impls.rs
```diff
@@ -116,7 +117,25 @@ where
...
+        self.view_block(BlockId::Hash(hash).into()).await
+    }
+
+    /// View the block from the network given the block reference to it.
+    pub async fn view_block(&self, block_ref: BlockReference) -> anyhow::Result<Block> {
```
austinabell, line 134: There doesn't seem to be much use in exposing the other APIs if this is still exported. Possibly an alternative would be to have `view_block_by_id` which takes in that BlockId and then `view_latest_block` which takes in the finality? 

Tough line to draw between function bloat and unergonomic functions in this case I think. Perhaps we could consider doing a builder pattern based on the finality/block id that can be re-used for all functions that reference specific blocks?

workspaces/src/types/mod.rs
```diff
@@ -214,3 +214,82 @@ impl From<AccessKey> for near_primitives::account::AccessKey {
...
+    }
+}
+
+/// Finality of a transaction or block in which transaction is included in. For more info
+/// go to the [NEAR finality](https://docs.near.org/docs/concepts/transaction#finality) docs
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum Finality {
+    /// Optimistic finality
+    None,
```
austinabell, line 242: Opinionated: Is none a bit misleading here? Why not Optimistic?

itegulov: I think this is just taken directly from nearcore to stay consistent: https://github.com/near/nearcore/blob/9fc9e318000a3a5cd40d34371ad1113fda6ee9ef/core/primitives/src/types.rs#L22-L32.

Probably not worth it to break the pattern, but I agree that the name is strange

workspaces/src/worker/impls.rs
ChaoticTempest: I should mention there is something I didn't include in `BlockReference` which is `SyncCheckpoint` since I didn't know whether we should expose that one quite yet or not, but not everything will be covered if we just have either finality or block_id related methods, so opting for the most common methods and having a more general one seems best in that case.

> Perhaps we could consider doing a builder pattern based on the finality/block id that can be re-used for all functions that reference specific blocks?

I'm not sure if everything would fit into the builder pattern mold going forward, but let's see for this case, we'd have:
```rust
worker.view_block().await
```
which can just be the default method, where we grab the latest block. Then we can have:
```
worker.view_block()
    .block_hash(...)
    // or
    .finality(Finality::Final)
    .await
```
I like since it does solve the function bloat problem, but would it make sense for us to have a builder just to add a single parameter?

workspaces/src/types/mod.rs
ChaoticTempest: Yeah, I'm not sure what's best to keep here, but I'm leaning towards doing `Finality::Optimistic`. In RPC https://docs.near.org/docs/api/rpc#using-finality-param, they specify "optimistic" which is definitely more user-friendly and they're breaking the pattern as well. @frol (tagging you since you had a commit on it) would you happen to know why it's called `Finality::None` instead of `Finality::Optimistic` in nearcore?

workspaces/src/worker/impls.rs
austinabell: Alternative maybe being:
```rust
worker
	// Optional
	.with_block_height()
	// Actual method
	.view_block()
```

So that things are consistent across methods, but I do accept that this might cause too much code bloat. I don't have a strong inclination of what will be best, up to you, just sharing suggestions so we don't build up technical debt

ChaoticTempest: yeah this one has so many ways to go about it 😅 

I think potentially, parameters can be added further into these queries so the builder pattern might just be the best use here, in terms of scalability of parameters. I see this already with `view_chunk` which has an extra `shard_id` parameter (which I need to add in a separate PR):
```
worker.view_chunk()
   .block_hash(...)
   .shard_id(0)
```

also, alternative would run into the similar issues of function bloat, especially on the surface level of worker. I like that the action has mainly been first with builders, but might not be a huge requirement there. How about anything that requires a `BlockReference`, we can just do a general builder like so which eliminates some code bloat:
```rust
enum ViewerMethod {
   Block,
   State(PrefixId),
   Code(AccountId),
   ...
}

struct Viewer {
   block_ref: BlockReference,
   method: ViewerMethod,
}
```

austinabell: Or one final option, just expose methods directly on `BlockReference` to make that parameter just easier to use to avoid builder patterns

```rust
impl BlockReference {
	fn code_hash(hash: CodeHash) -> Self {..}
	fn height(..) -> Self {..}
	fn latest() ..
	// idk about this one, maybe separate functions for doomslug and final would be better to avoid exposing finality
	fn latest_with_finality(..)
}
```

And then maybe even obfuscate BlockReference enum type to streamline the API since this would only be used for input (correct?)

ChaoticTempest: yeah it's only used with input. Let's go with final option then. Wouldn't add too much code either

austinabell: I guess with this we might want to come up with consistent semantics if this is reused. Is it something `<method>` with the latest default and `<method>_with_reference` for customizing? This is definitely opinionated

ChaoticTempest: ahh that part is probably gonna put us back into builders again if we don't want function bloat 😅, considering `<method>_with_unknown_future_parameter` could happen. It's also not API breaking if we were to switch stuff like `view_code` to include the builder pattern:
```
worker.view_code(id)
   .block_ref(...)
   .await
```

So maybe a combination of both would be best? Or maybe we just accept function bloat and call it a day haha

workspaces/src/types/mod.rs
frol: I was only in charge of exposing those variants via RPC interface and so you can see I turned it into "optimistic" for the public interfaces, and I never questioned the reasoning for this naming internally. Ask @bowenwang1996 who implemented it in https://github.com/near/nearcore/pull/2127, but all-in-all, I think the answer will be "it just happened this way for no specific reason", and given it was not part of the public interface back in the days, we never questioned it. Probably, it is a good time to rename those variants for the public interfaces needs.

ChaoticTempest:
close this PR in favor of https://github.com/near/workspaces-rs/pull/218

Issue #132, opened 2022-05-10
Is there a reason that CallExecutionDetails.try_into_bytes is private?


willemneal:
I need to access the raw bytes, but seem to have to reimplement this to access them:

https://github.com/near/workspaces-rs/blob/50206b6ac65b8e6b225d848862f8dc25c8b4d2b2/workspaces/src/result.rs#L84-L92

willemneal:
In fact I can't implement outside the crate this because the `status` field is also private.

austinabell:
that's very valid. This function probably should be exposed (not sure under that function name, though)

ChaoticTempest:
definitely an assumption that only Json or Borsh deserialization would be needed. Not sure what other name to give it though, maybe just `bytes()`?

austinabell:
> definitely an assumption that only Json or Borsh deserialization would be needed. Not sure what other name to give it though, maybe just `bytes()`?

Another candidate: `raw_bytes` or `return_bytes`

ChaoticTempest:
Went with `raw_bytes` in https://github.com/near/workspaces-rs/pull/133

PR #133, opened 2022-05-10
chore: expose raw_bytes API

ChaoticTempest:
addresses https://github.com/near/workspaces-rs/issues/132

### DIFF
--- a/workspaces/src/result.rs
+++ b/workspaces/src/result.rs
@@ -81,6 +81,13 @@ impl CallExecutionDetails {
         borsh::BorshDeserialize::try_from_slice(&buf).map_err(Into::into)
     }
 
+    /// Grab the underlying raw bytes returned from calling into a contract's function.
+    /// If we want to deserialize these bytes into a rust datatype, use [`CallExecutionDetails::json`]
+    /// or [`CallExecutionDetails::borsh`] instead.
+    pub fn raw_bytes(&self) -> anyhow::Result<Vec<u8>> {
+        self.try_into_bytes()
+    }
+
     fn try_into_bytes(&self) -> anyhow::Result<Vec<u8>> {
         let result: &str = match self.status {
             FinalExecutionStatus::SuccessValue(ref val) => val,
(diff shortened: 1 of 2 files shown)


workspaces/src/result.rs
```diff
@@ -81,6 +81,13 @@ impl CallExecutionDetails {
         borsh::BorshDeserialize::try_from_slice(&buf).map_err(Into::into)
     }
 
+    /// Grab the underlying raw bytes returned from calling into a contract's function.
+    /// If we want to deserialize these bytes into a rust datatype, use [`CallExecutionDetails::json`]
+    /// or [`CallExecutionDetails::borsh`] instead.
+    pub fn raw_bytes(&self) -> anyhow::Result<Vec<u8>> {
+        self.try_into_bytes()
```
austinabell, line 88: Could have just renamed try_into_bytes?

workspaces/src/result.rs
```diff
@@ -81,6 +81,13 @@ impl CallExecutionDetails {
         borsh::BorshDeserialize::try_from_slice(&buf).map_err(Into::into)
     }
 
+    /// Grab the underlying raw bytes returned from calling into a contract's function.
+    /// If we want to deserialize these bytes into a rust datatype, use [`CallExecutionDetails::json`]
+    /// or [`CallExecutionDetails::borsh`] instead.
+    pub fn raw_bytes(&self) -> anyhow::Result<Vec<u8>> {
```
austinabell, line 87: I wonder if raw_bytes could be ambiguous with the raw response bytes rather than just the return bytes. Maybe another opinion on this naming might be worthwhile for something like `return_bytes`

austinabell (approved):

workspaces/src/result.rs
ChaoticTempest: @volovyk-s @itegulov what do you think would be better naming here for grabbing a contract's executed function return value as bytes?

willemneal: I read the current API for the other two calls as `d.as_json()`.  So `bytes` would fit the pattern.

itegulov: The problem with `return_bytes` is that it reads as a verb + noun while it is actually adj + noun. I think `raw_bytes` makes sense to me

austinabell: Yeah true, but it was intended as an adjective :D as in these are the "return value" bytes, but yeah I see the confusion here

ChaoticTempest: Alright, guess that's settled. `raw_bytes` it is then

PR #134, opened 2022-05-11
chore: bump sandbox version default for SDK change

austinabell:
commit is `1.26.0-rc.2` for now -- tested it works against https://github.com/near/near-sdk-rs/pull/742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [unreleased]
 
+### Changed
+- Updated default sandbox version to `1.26.0-rc.2` of nearcore
+
 ## [0.2.1] - 2022-04-12
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [unreleased]
 
 ### Changed
-- Updated default sandbox version to `1.26.0-rc.2` of nearcore
+- Updated default sandbox version to `97c0410de519ecaca369aaee26f0ca5eb9e7de06` commit of nearcore to include 1.26 protocol changes
 
 ## [0.2.1] - 2022-04-12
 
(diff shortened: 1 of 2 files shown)


ChaoticTempest (approved):

itegulov (commented):
While this works for Linux-x86_64, the same binary is not available for Darwin-x86_64: https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore/Darwin-x86_64/1.26.0/57b6ac92b0e77deade8a6eef1a322272511aa5de/near-sandbox.tar.gz

This is the issue I was talking about the other day

ChaoticTempest:
hmmm, I wonder if the `mac-binary.yml` regex pattern isn't catching `-rc.2` naming:
```
    tags:
      - '[0-9]+.[0-9]+.[0-9]+.*'
```

it's really hard to test GitHub Actions. A test for a test

austinabell:
> hmmm, I wonder if the `mac-binary.yml` regex pattern isn't catching `-rc.2` naming:
> 
> ```
>     tags:
>       - '[0-9]+.[0-9]+.[0-9]+.*'
> ```
> 
> it's really hard to test GitHub Actions. A test for a test

Nah the regex matches, not sure here. Maybe has to do with the fact that the tagged commit isn't on master?

ChaoticTempest:
> I think it might be because the regex is in ''

yeah that could be it and if so, we should probably get in a fix before 1.26 complete rolls out

austinabell:
> > I think it might be because the regex is in ''
> 
> yeah that could be it and if so, we should probably get in a fix before 1.26 complete rolls out

Not sure though, actions docs are inconsistent, but can see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags shows no `'`. Or the issue is just that the tag is applied after push and then the workflow is not run -- more likely?

ChaoticTempest:
yeah let's just roll this one in. We were previously using a non-stable commit too. We can fix the regex in the meantime

PR #135, opened 2022-05-11
fix: error if sandbox process has already started

austinabell:
Not sure if this is intentional, opening PR more as a question of why this is the case that the process handle can get overwritten here

### DIFF
--- a/workspaces/src/network/server.rs
+++ b/workspaces/src/network/server.rs
@@ -25,6 +25,10 @@ impl SandboxServer {
     }
 
     pub fn start(&mut self) -> anyhow::Result<()> {
+        if self.process.is_some() {
+            anyhow::bail!("Sandbox server already started");
+        }
+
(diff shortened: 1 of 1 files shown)


workspaces/src/network/server.rs
```diff
@@ -25,6 +25,10 @@ impl SandboxServer {
     }
 
     pub fn start(&mut self) -> anyhow::Result<()> {
+        if self.process.is_some() {
+            anyhow::bail!("Sandbox server already started");
+        }
```
ChaoticTempest, line 30: huh, what case were you hitting where `start` was being called multiple times? Still good to have this in here just in case though

ChaoticTempest (approved):

workspaces/src/network/server.rs
austinabell: No, but reading the logic it seemed like it could be a hidden bug if logic changed external to this method

Issue #130, opened 2022-05-02
Can't use `SecretKey`


NinoLipartiia4ire:
Can't create a variable of type `SecretKey` because module `types` is private. Probably users should be able to do that in order to use some of public methods, e.g., for adding keys to the new account.

ChaoticTempest:
@NinoLipartiia4ire what kind of `SecretKey` creation methods are you thinking would be useful here? I'm considering adding a `FromStr` impl but unsure what other use cases there could be

itegulov:
I think @NinoLipartiia4ire means that they can't do the following pattern since there is no way to import `SecretKey` type:

```rust
let sk: SecretKey = worker.dev_generate().await.1;
```

While you can omit the type in this particular case and use the variable locally, you won't be able to pass it around to other functions. Does this make sense?

ChaoticTempest:
oh wait, I assumed we already had exposed the module `types` back in 0.2, but that's not the case. I was thinking we could've done `workspaces::types::SecretKey` this whole time. I'll update my PR to include it

PR #139, opened 2022-05-12
chore: add secret key creation methods and added new KeyType type

ChaoticTempest:
addresses https://github.com/near/workspaces-rs/issues/130

`SecretKey` now has two new creation methods:
- `from_random`
- `from_str`

~Somewhat breaking API if anyone used `SecretKey::from_seed`, since this PR swapped `near_crypto::KeyType` for `workspaces::types::KeyType`~

edit: nevermind about breaking API, `types` module was never exposed

### DIFF
--- a/workspaces/src/types/mod.rs
+++ b/workspaces/src/types/mod.rs
@@ -9,7 +9,7 @@ use std::fmt;
 use std::path::Path;
 
 pub use near_account_id::AccountId;
-pub(crate) use near_crypto::{KeyType, Signer};
+pub(crate) use near_crypto::Signer;
 use near_primitives::logging::pretty_hash;
 use near_primitives::serialize::{from_base, to_base};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


workspaces/src/types/mod.rs
```diff
@@ -28,25 +28,79 @@ pub type Balance = u128;
...
+}
+
+impl From<near_crypto::KeyType> for KeyType {
+    fn from(key_type: near_crypto::KeyType) -> Self {
+        match key_type {
+            near_crypto::KeyType::ED25519 => Self::ED25519,
+            near_crypto::KeyType::SECP256K1 => Self::SECP256K1,
+        }
+    }
+}
```
austinabell, line 55: Not 100% sure we want this impl to be public (for maintenance reasons) is there any reason someone might want to convert to this rather than just having it as a helper function?

workspaces/src/types/mod.rs
```diff
@@ -28,25 +28,79 @@ pub type Balance = u128;
...
+        Self(near_crypto::SecretKey::from_random(key_type.into()))
+    }
+}
+
+impl std::str::FromStr for SecretKey {
+    // FIXME: deferred specific error impl to https://github.com/near/workspaces-rs/issues/100
+    type Err = anyhow::Error;
+
+    fn from_str(value: &str) -> anyhow::Result<Self> {
+        Ok(Self(near_crypto::SecretKey::from_str(value)?))
```
austinabell, line 103: noting that we will have to be careful on updating this dep that the semantics don't change. Not sure how stable this API is

austinabell (approved):

austinabell (approved):

workspaces/src/types/mod.rs
austinabell: What you changed it to lgtm, we can always add it later, but a more breaking change to remove and this way the API is more streamlined and harder to misuse

Issue #142, opened 2022-06-01
Validator stake pool boostrap for local sandbox


bharath-123:
Hi all,

I am using the latest workspaces-rs version:
workspaces = { git = "https://github.com/near/workspaces-rs" }

I am trying to write contracts which will integrate with the validator stake pool contracts. With unstaking and rewards accrual taking an epoch, its not easy to test these contracts on testnet. The epoch forward feature in the local sandbox is very useful in such cases. it would be great if the sandbox can be bootstrapped with a validator stake pool too. That would make it very easy to test stake pool integrations. 

Currently we would have to deploy a stake pool with a public key. I don't see a straightforward way to get the validator public key from the workspace-rs library. but i may be wrong. This process is very manual. Boostrapping it can make the dev experience nicer.

Since stake pools is a core part of the eco system, there will be a lot of integrations with this contract. i definitely see value in it.

ChaoticTempest:
I see, thanks for bringing this up! I'll make the keys available for usage then. The credentials are already being exposed to the users to save as a file, but that's quite involved for users to grab just to send the bytes into a transaction.

Also, if I understand what you're saying about bootstrapping correctly, having the staking pool contract inside workspaces library itself would make it bloated right now. It's best if we have it at a higher level library to facilitate that, just so that workspaces-rs can focus purely on testing and automation. But definitely, this is good feedback that it would make it easier to plugin such libraries to test and play around with it later on

bharath-123:
hey @ChaoticTempest Thanks for the quick turnover! I understand your concern about the bloat. But I also do feel that maybe boot strapping the sandbox with core contracts could be a good idea. Even if the user has to provide the wasm, some functionality to setup a stake pool contract and a complete validator would be really useful.

ChaoticTempest:
hmm, well most of the functionality (after exposing the keys) should be available for users to make use of to deploy the staking contract. workspaces-rs is already running a node underneath and that will pretty much be your validator node. We just gotta deploy the staking contract and initialize it with our data. It'd look something like this if I'm understanding the staking contract correctly:
```rust
let worker = workspaces::sandbox().await?;

let pk = worker.root_account().secret_key().public_key();
let owner = worker.dev_create_account().await?;
let staking_contract = worker.dev_deploy("./path/to/staking-contract.wasm").await?;

owner.call(&worker, staking_contract.id(), "new")
    .args_json(json!({
         "owner_id": owner.id(),
         "stake_public_key": pk,
         ... 
[...]

PR #144, opened 2022-06-02
chore: Expose account keys

ChaoticTempest:
Addresses https://github.com/near/workspaces-rs/issues/142

Exposes the account keys. Might be a little weird that it's exposed just as a `SecretKey` type instead of something like a `KeyPair` type. This is due to `SecretKey` being a `KeyPair` itself containing the `PublicKey`. WDYT? Is the return type good enough just as `SecretKey` or should we do something like `(PublicKey, SecretKey)`?

### DIFF
--- a/workspaces/src/types/account.rs
+++ b/workspaces/src/types/account.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 
 use near_primitives::views::AccountView;
 
-use crate::types::{AccountId, Balance, InMemorySigner};
+use crate::types::{AccountId, Balance, InMemorySigner, SecretKey};
 use crate::{CryptoHash, Network, Worker};
 
 use crate::operations::{CallTransaction, CreateAccountTransaction, Transaction};
@@ -146,6 +146,11 @@ impl Account {
         crate::rpc::tool::write_cred_to_file(&savepath, &self.id, &self.signer.0.secret_key);
 
         Ok(())
+    }
+
+    /// Get the keys of this account. The public key can be retrieved from the secret key.
+    pub fn keys(&self) -> SecretKey {
+        SecretKey(self.signer.0.secret_key.clone())
     }
 }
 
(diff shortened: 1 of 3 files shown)


workspaces/src/types/account.rs
```diff
@@ -147,6 +147,11 @@ impl Account {
 
         Ok(())
     }
+
+    /// Get the keys of this account. The public key can be retrieved from the secret key.
+    pub fn keys(&self) -> SecretKey {
+        SecretKey(self.signer.0.secret_key.clone())
+    }
```
itegulov, line 154: I think it is fine to only return `SecretKey`, but I would argue that `keys` is not a good name in this case because of plurality. Maybe just `key` or `secret_key` while leaving the clarification that public key can be retrieved from the secret key?

austinabell: Also, noting it's probably best if we keep this representation internally, and then this method can return `&SecretKey`. Cloning implicitly feels weird

ChaoticTempest: Gotcha, I'll re-work how `InMemorySigner` is represented internally but shouldn't be too big of a deal to expose it as a `&SecretKey`

ChaoticTempest: alright should be good. Ready to be re-reviewed whenever you guys get the chance

austinabell (approved):
Not completely sure how this `Signer` type. I suppose it has to do with creating TLAs? Maybe you can clarify how you plan to fit this in cleanly?

ChaoticTempest:
> Not completely sure how this Signer type.

Reading in between the lines, but I'm guessing you meant to say how does the `Signer` type work? `Signer` is mainly used to sign transactions and hold keys, but good catch since this detail doesn't necessarily need to be exposed to the user. I'll make a ticket to evaluate if we can clean this up in another PR

Issue #100, opened 2022-04-04
Error Handling


matklad:
Continuing discussion from:

[https://github.com/near/nearcore/pull/6367#issuecomment-1069525604](https://github.com/near/nearcore/pull/6367#issuecomment-1069525604)

TL;DR:

My understanding is that today worspaces expose nearcore-internal errors as `anyhow::Error`. That is, it's possible to get the original error of of `anyhow::Error` by downcasting.

This is a backwards-compatability hazard – at nearcore, we are free to change the errors however we like, but this now can break clients of workspaces.

In terms of what to do about this – if we don't expect the clients to actually inspect the errors by downcasting, than this is probably fine. Like, this is still something to be fixed in a top-notch library, but it's unlikely to cause significant trouble.

If we do expect the clients to actually downcast the error to handle it, than that I think is suboptimal. `anyhow` _can_ be use for selective error handling, but it's not ergonomic or reliably (as one forgoes static checks). If the users indeed might want to handle errors originating from workspaces, I suggest using the error kind pattern instead:

<pre>pub struct WorkspaceError 

 { // Opaque representation, repr: Whatever } 

impl WorkspaceError 

 { pub fn kind(&self) -> WorkspaceErrorKind; } 

#<span class="error">[non_exhaustive]</span>
 pub enum WorkspaceErrorKind 

 { SandboxNotRunning, ServerError, ... } 

</pre>

It strikes a nice balance between allowing the caller to selectively handle errors, and allowing implementer making semver-compatible changes.

*   * *

One related bit is that we have this `TxExecutionError` "error" type here:

[https://github.com/near/nearcore/blob/cf6c3177dd0b0bc883038d5c320d92197dd9f315/core/primitives/src/views.rs#L1017](https://github.com/near/nearcore/blob/cf6c3177dd0b0bc883038d5c320d92197dd9f315/core/primitives/src/views.rs#L1017)

I think is would be wrong to treat this as an error at the workspace level – `TxExectuionError` is not some kind of transient environmental error. It's a normal outcome of a transaction which failed to apply. It's a value type with a specific serialization format, etc. I think the signature we want for `transct` method which submits transaction and gets it should have roughly the same shape:

<pre>fn transact(&self, tx: Transaction) -> anyhow::Result

pub struct ExecutionOutcome 

 { pub logs: ... pub gas_burnt: ... pub error: Option , } 

 </pre>

[...]

PR #149, opened 2022-06-14
feat: error handling

ChaoticTempest:
This PR resolves https://github.com/near/workspaces-rs/issues/100

Naming is always challenging, so let me know if there's any naming in the errors that are confusing.

Also, note that I didn't change any of the `compile_project` related errors since I didn't wanna snowball this one -- I'll change those later after this PR lands

## Errors layout:
The surface `Error` enum type is used specifically to pinpoint the error a user might want. The underlying errors such as `ParseError` and `RpcError` hides the details being exposed from dependencies so that we can represent the data however we want if the data changes from these types on nearcore side in the future.

I might've also messed up how `thiserror` library is intended to be used since I used it on the `ErrorKind` type so I could inline the error message into the variant itself without having to write the `Display` impl.

@matklad, let me know what you think if you have time


workspaces/src/network/server.rs
```diff
@@ -37,9 +38,14 @@ impl SandboxServer {
 
         // Remove dir if it already exists:
         let _ = std::fs::remove_dir_all(&home_dir);
-        sandbox::init(&home_dir)?.wait()?;
+        sandbox::init(&home_dir)
+            .map_err(|e| Error::SandboxUnknownError(e.to_string()))?
+            .wait()
+            .map_err(Error::IoError)?;
```
itegulov, line 44: Have you considered a separate error type for sandbox initialization issues?

workspaces/src/rpc/client.rs
```diff
@@ -366,9 +378,11 @@ impl Client {
         result
     }
 
-    pub(crate) async fn wait_for_rpc(&self) -> anyhow::Result<()> {
+    pub(crate) async fn wait_for_rpc(&self) -> Result<(), RpcError> {
         let timeout_secs = match std::env::var("NEAR_RPC_TIMEOUT_SECS") {
-            Ok(secs) => secs.parse::<usize>()?,
+            // hard fail on not being able to parse the env var, since this isn't something
+            // the user should handle with the library.
+            Ok(secs) => secs.parse::<usize>().unwrap(),
```
itegulov, line 385: Shouldn't we at least supply an error message here? It would be confusing to see workspaces panic with something like "ABC is not a number"

workspaces/src/rpc/client.rs
```diff
@@ -229,22 +236,27 @@ impl Client {
...
+    ) -> Result<BlockView, RpcError> {
         let block_reference = block_id
             .map(Into::into)
             .unwrap_or_else(|| Finality::None.into());
 
         let block_view = self
             .query(&methods::block::RpcBlockRequest { block_reference })
-            .await?;
+            .await
+            .map_err(|e| RpcErrorKind::QueryBlockFailure.with_repr(e.into()))?;
```
itegulov, line 259: Do you have some kind of criteria why some of the calls have their own Failure type (e.g. `QueryBlockFailure`), but others use a generic one (`QueryFailure`)?

workspaces/src/rpc/patch.rs
```diff
@@ -81,7 +82,7 @@ impl<'a, 'b> ImportContractTransaction<'a, 'b> {
     }
 
     /// Process the trannsaction, and return the result of the execution.
```
itegulov, line 84: typo
```suggestion
    /// Process the transaction, and return the result of the execution.
```

workspaces/src/rpc/tool.rs
```diff
@@ -38,19 +39,23 @@ pub(crate) async fn url_create_account(
...
+            serde_json::to_vec(&serde_json::json!({
+                "newAccountId": account_id,
+                "newAccountPublicKey": pk,
+            }))
+            .map_err(BytesError::SerdeError)?,
+        )
         .send()
-        .await?;
+        .await
+        .map_err(|e| Error::Other(Box::new(e)))?;
```
itegulov, line 58: Would not this classify as RpcError?

workspaces/src/result.rs
```diff
@@ -68,45 +65,54 @@ impl CallExecutionDetails {
...
-            FinalExecutionStatus::Started => anyhow::bail!("Transaction still being processed."),
-            _ => (),
-        };
+    fn try_into_result(self) -> Result<Self> {
+        self.try_into_success_value()?;
         Ok(self)
     }
 
-    pub(crate) fn from_outcome(outcome: FinalExecutionOutcomeView) -> anyhow::Result<Self> {
+    pub(crate) fn from_outcome(outcome: FinalExecutionOutcomeView) -> crate::result::Result<Self> {
```
itegulov, line 115: Unnecessary full path here too. There seem to be a few of those around, not sure if that was intentional, lmk your reasoning if it is

workspaces/src/error/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error("parse error from {0}")]
+    ParseError(#[from] ParseError),
+    #[error("bytes error from {0}")]
+    BytesError(#[from] BytesError),
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error>),
+}
+
+unsafe impl Sync for Error {}
+unsafe impl Send for Error {}
```
matklad, line 40: This is definitely very wrong and unsound. You need `Other(#[from] Box<dyn std::error::Error + Send + Sync>),` and a test to check that the `Error` is sync+send instead.

workspaces/src/error/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    IoError(#[from] std::io::Error),
+    #[error("account error from {0}")]
+    AccountError(String),
+    #[error("parse error from {0}")]
+    ParseError(#[from] ParseError),
+    #[error("bytes error from {0}")]
+    BytesError(#[from] BytesError),
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error>),
+}
```
matklad, line 37: What is the motivation for *not* making this an opaque error types? `SandboxUnknownError(String),` and `Other(#[from] Box<dyn std::error::Error>)` as a publicly visible API strike me as highly unexpected. 

I would expect something shaped roughtly like reqwest error: 

https://docs.rs/reqwest/latest/reqwest/struct.Error.html

workspaces/src/error/rpc.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    pub(crate) fn from_msg(kind: RpcErrorKind, msg: &'static str) -> Self {
+        Self::from_repr(kind, anyhow::anyhow!(msg).into())
+    }
+
+    /// Get the kind of error that occurred in the RPC service.
+    pub fn kind(&self) -> &RpcErrorKind {
+        &self.kind
+    }
```
matklad, line 92: kind is usually deliberately made a value-less copy enum,

```suggestion
    pub fn kind(&self) -> RpcErrorKind {
        self.kind
    }
```

workspaces/src/error/rpc.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Get the underlying error message from respective error. This can be
+    /// empty to signify no meaningful error message is present.
+    pub fn err_msg(&self) -> String {
```
matklad, line 96: The right public API for to_string conversion is `Display`

workspaces/src/operations.rs
```diff
@@ -231,14 +234,14 @@ impl<'a, 'b, T: Network> CallTransaction<'a, 'b, T> {
     /// Similiar to `args`, specify an argument that is JSON serializable and can be
     /// accepted by the equivalent contract. Recommend to use something like
     /// `serde_json::json!` macro to easily serialize the arguments.
-    pub fn args_json<U: serde::Serialize>(mut self, args: U) -> anyhow::Result<Self> {
+    pub fn args_json<U: serde::Serialize>(mut self, args: U) -> Result<Self, BytesError> {
         self.function = self.function.args_json(args)?;
         Ok(self)
     }
```
matklad, line 240: Consider making this method infalible and using `reqwest` approach of postponing the error to the point where the request is made: 

https://docs.rs/reqwest/latest/src/reqwest/async_impl/request.rs.html#442-455

workspaces/src/operations.rs
```diff
@@ -276,11 +279,12 @@ impl<'a, 'b, T: Network> CallTransaction<'a, 'b, T> {
                 self.function.deposit,
             )
             .await
+            .map_err(crate::error::Error::from)
             .and_then(CallExecutionDetails::from_outcome)
```
matklad, line 283: If I am reading this correctly, this'll convert faield transactions into errors, erasing all information except for status. I think this is wrong: "the RPC failed due to IO error" and "transaction execution itself failed due to contract returning with error" seems to be very dissimilar error conditions. 

This should be `map(CallExecutionDetails::from)` rather than `and_then(from_outcome)`

matklad (commented):
I have one big design consideration: what is the motivation for *not* making the error opaque error type? To me, for a sprawling, high-level library like workspaces an opaque error types seems like the most obvious solution, by a big marging. 

Exposing internal details:

* constraints us semver-wise
* confuses the user (which error variants should they handle)
* gives "this is not well-designed" feeling overall: why there's "UnknownError" but also "Other"? why "BytesError" are grouped, but sandbox errors aren't?

austinabell (commented):
Going to have a deeper look into this later looking into those conversions, but seems like it's trending in the right direction!

austinabell:
> I have one big design consideration: what is the motivation for _not_ making the error opaque error type? To me, for a sprawling, high-level library like workspaces an opaque error types seems like the most obvious solution, by a big margin.

I believe the motivation is that certain errors may want to be handled differently than others (for example some internal error vs some network failure) and obfuscating details or forcing converting into a string and matching a pattern might not be the best solution. Maybe there is a way we can expose details about the types of errors without exposing everything as is with this PR, because this currently definitely is more of a backwards-compatible issue than the anyhow errors previously.

ChaoticTempest:
> I have one big design consideration: what is the motivation for not making the error opaque error type? To me, for a sprawling, high-level library like workspaces an opaque error type seems like the most obvious solution, by a big margin.

What @austinabell mentioned is why I went with this top-level error enum. It's just unwieldy for a user if they want to handle a specific class of errors which I feel would be a common case. Not all classes of errors would fit in with one kind of data representation either. The reqwest example you linked out to just exposes representation relating to network, but these errors in workspaces can be more than just those, so the top-level enum was what I thought could encapsulate this, where each variant is its own representation of the underlying data.

> Maybe there is a way we can expose details about the types of errors without exposing everything as is with this PR, because this currently definitely is more of a backwards-compatible issue than the anyhow errors previously.

I intentionally left some of the errors as unwrapped and bloating top-level to get some feedback, but I can wrap up them into their own `ErrorKind`/`ErrorRepr` type

workspaces/src/error/mod.rs
ChaoticTempest: Definitely need to add a test here, but since workspaces needs to run in a runtime, they need to be sent across threads which is why `Error` here needs to be `Sync + Send`:
```rust
workspaces::sandbox().await?;
                           ^ `(dyn std::error::Error + 'static)` cannot be sent between threads safely
```

Any alternatives to this then?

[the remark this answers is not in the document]
ChaoticTempest: They're required since we'd get this error otherwise when making calls across threads:
```rust
workspaces::sandbox().await?;
                           ^ `(dyn std::error::Error + 'static)` cannot be sent between threads safely
```
Any suggestions to avoid this?

workspaces/src/error/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum BytesError {
+    #[error("serde error: {0}")]
+    SerdeError(#[from] serde_json::Error),
+    #[error("borsh error: {0}")]
+    BorshError(std::io::Error),
+    #[error("failed to decode to base64 due to {0}")]
+    DecodeBase64Error(#[from] base64::DecodeError),
+}
```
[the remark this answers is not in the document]
ChaoticTempest: I originally had `SerializationError`, but went with `BytesError` since it seems more general. But Base64 encode/decode is pretty much serialization as well; which is still pretty much serialization overall, so let's just change it to `SerializationError` for now

workspaces/src/result.rs
```diff
@@ -265,18 +275,18 @@ impl ExecutionOutcome {
         )
     }
 
-    /// Converts this [`ExecutionOutcome`] into a Result type, where the failure is converted
-    /// to an [`anyhow::Error`] object which can be downcasted later.
-    pub fn into_result(self) -> anyhow::Result<ValueOrReceiptId> {
+    /// Converts this [`ExecutionOutcome`] into a Result type to match against whether the
+    /// particular outcome has failed or not.
+    pub fn into_result(self) -> crate::result::Result<ValueOrReceiptId> {
```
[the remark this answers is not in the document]
ChaoticTempest: woops, this was because I had different `Error` types and I kept changing their names as I went on developing. This was left over due to me regex replacing some `Result` with the full path

workspaces/src/network/server.rs
ChaoticTempest: good point, added it

workspaces/src/rpc/client.rs
ChaoticTempest: ahh, that's a good point! They don't really make much of a difference. They can all just be `QueryFailure`s, unless you think they should all be more explicit like `QueryBlockFailure`? Changed to `QueryFailure` in latest commit

workspaces/src/rpc/tool.rs
ChaoticTempest: nice catch! This definitely belongs with RPC

workspaces/src/error/mod.rs
ChaoticTempest: pretty much in the general comment of https://github.com/near/workspaces-rs/pull/149#issuecomment-1156908342 where we have too many different sets of errors that is not intuitive to work with if we just muddle the waters here, IMO

workspaces/src/operations.rs
ChaoticTempest: this was to force users to handle the failed transaction instead of ignoring them, but definitely loses info. I then learned about `#[muse_use]` a while back and went into this new direction with this PR: https://github.com/near/workspaces-rs/pull/150 to handle this case specifically

workspaces/src/error/rpc.rs
```diff
@@ -0,0 +1,102 @@
...
+        Self::from_repr(kind, anyhow::anyhow!(msg).into())
+    }
+
+    /// Get the kind of error that occurred in the RPC service.
+    pub fn kind(&self) -> &RpcErrorKind {
+        &self.kind
+    }
+
+    /// Get the underlying error message from respective error. This can be
+    /// empty to signify no meaningful error message is present.
```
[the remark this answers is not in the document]
ChaoticTempest: So with Aleksey's recommendation of having the public API for to string conversion being just `Display`, this function is no longer being exposed

workspaces/src/network/testnet.rs
```diff
@@ -88,7 +90,7 @@ impl TopLevelAccountCreator for Testnet {
                     logs: Vec::new(),
                     receipt_ids: Vec::new(),
                     gas_burnt: 0,
-                    executor_id: "testnet".parse()?,
+                    executor_id: "testnet".parse().unwrap(),
```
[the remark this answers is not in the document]
ChaoticTempest: yup, but now that you point it out, since this is in testnet, it's not covered by current tests so will add a new one

workspaces/src/rpc/client.rs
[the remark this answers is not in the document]
ChaoticTempest: hmm, normally for user input errors, wouldn't we want to hardfail though? Since that gives immediate feedback, and I just see this being a case that won't get handled at all. It just feels weird when we go to handle it, since it would just be used for potentially cleanup. But then again, for things like dev console, we'd want to package errors as cleanly as possible. I do see value with having it in that case. What would be a good name for the class of errors of these? `InputError`? Or should we put it in something like `ParseError`?

workspaces/src/error/mod.rs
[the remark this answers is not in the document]
ChaoticTempest: ah nevermind, I'm just being dumb, since they get auto-impl'd as long as the members are sync/send

matklad:
>if they want to handle a specific class of errors which I feel would be a common case.

What are some specific examples here? If the API forces the caller to handle a *subset* of the errors, it feels like a footgun. It's better if the cases which need handling can be expressed separately, such that all the errors are unconditionally propagated.

ChaoticTempest:
Something like the following is what I was leaning into:
```rust
let result = worker.call(...).transact().await;
if let Err(error) = result {
   match error {
       ExecutionError(err: String) => { // do something with err }
       SerializationError(err) => { // do something with err }
       other => Err(other)?,  // propagate errors we don't care about
   }
}
```
where the user can match to a specific class of errors such as an `ExecutionError` vs a `SerializationError`, which can come from that single function call into a contract. 

> If the API forces the caller to handle a subset of the errors, it feels like a footgun. It's better if the cases which need handling can be expressed separately, such that all the errors are unconditionally propagated.

I feel the above example represents that we can handle errors separately unless I'm missing something with your statement.

matklad:
Yeah, I understand that making error an enum allows matching on the error variants. What I don't understand is *why* would the user need to do that? That is what that "something" would be? 

My prior is that matching on errors should be exceedingly rare, and that we shouldn't optimize for this use-case.

ChaoticTempest:
Ah, I see what you're getting at. For something like Serialization/Parse error, we don't need to explicitly handle that case since they're more informative for debugging purposes (in which case I'll just get rid of those for something more opaque), but errors like `RpcError` can be useful for determining exact causes and handling them can be useful for when one wants to make an automation script out of workspaces instead of just testing. I guess in that case, we can just eliminate the `Error` enum down to just `ExecutionError`, `RpcError`, and something like `Other` to capture the rest. WDYT?

matklad:
I suggest taking a look at how azure sdk handles errors: https://docs.rs/azure_core/0.3.0/azure_core/error/index.html

Imo, that's the right pattern: 

* opaque error struct
* with a `kind` method
* which returns very little data (mostly a field-less enum, empathically _not_ the actual internal representation)
* with a special case for HTTP errors

workspaces/src/error/impls.rs
```diff
@@ -0,0 +1,199 @@
...
+            ErrorRepr::Message { kind, .. } => kind,
+            ErrorRepr::Custom { kind, .. } => kind,
+            ErrorRepr::Full { kind, .. } => kind,
+        }
+    }
+
+    /// Consumes the `Error`, returning its inner error (if any).
+    ///
+    /// If this [`Error`] was constructed via [`Error::custom`] or [`Error::full`]
+    /// then this function will return [`Some`], otherwise it will return [`None`].
```
austinabell, line 79: This isn't a valid comment anymore

workspaces/src/error/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+/// A list specifying general categories of NEAR workspace error.
+#[derive(Clone, Debug, PartialEq, thiserror::Error)]
+#[non_exhaustive]
+pub enum ErrorKind {
+    /// An error occurred while performing an RPC request.
+    #[error("Rpc({0})")]
+    Rpc(#[from] RpcErrorCode),
+    /// An error occurred while processing a transaction.
+    #[error("Execution")]
+    Execution,
```
austinabell, line 17: Intuitively seems like someone might want to check that the execution error was a specific string/substring. We can always keep the API minimal and add later, though

workspaces/src/operations.rs
```diff
@@ -52,16 +56,22 @@ impl<'a> Function<'a> {
     /// Similiar to `args`, specify an argument that is JSON serializable and can be
     /// accepted by the equivalent contract. Recommend to use something like
     /// `serde_json::json!` macro to easily serialize the arguments.
-    pub fn args_json<U: serde::Serialize>(mut self, args: U) -> anyhow::Result<Self> {
-        self.args = serde_json::to_vec(&args)?;
-        Ok(self)
+    pub fn args_json<U: serde::Serialize>(mut self, args: U) -> Self {
```
austinabell, line 59: it doesn't make much sense to me that this method would not return a result anymore and keep the error until the call is executed. It seems to me that it should be either:
- Keep the signature as is, do serialization when args_x is called
- Make `Function` generic on the args type, serialize when function is executed

What is the motivation for having this change included with the error handling?

workspaces/src/rpc/client.rs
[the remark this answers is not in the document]
austinabell: The issue is that this might be used other than just with a test, and this wait_for_rpc would panic when someone might want to gracefully handle errors and re-start the sandbox or something else. I don't see how this would want to be handled differently than the error below.

At least, Daniyar's suggestion of adding an error message with this feels necessary.

workspaces/src/error/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Error type that workspaces will make use of for all the errors
+/// returned from this library
+#[derive(Debug)]
+pub struct Error {
+    repr: ErrorRepr,
+}
+
+#[derive(Clone, Debug, PartialEq, thiserror::Error)]
+pub enum SandboxErrorCode {
```
matklad, line 59: missing `#[non_exhaustive]`

workspaces/src/error/mod.rs
matklad: I would say it would be wrong to use `ErrorKind::Execution` to get details about execution. tx execution outcome is not a `Result<T, E>` -- even a failed transaction spends gas. The API should push the users who wish to inspect execution errors to use `CallExecutionDetails` directly. If the user need to see what the error was for debugging purposes, there's `fmt::Display` for that

workspaces/src/error/impls.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn from(error: base64::DecodeError) -> Self {
+        Self::custom(ErrorKind::DataConversion, error)
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(error: serde_json::Error) -> Self {
+        Self::custom(ErrorKind::DataConversion, error)
+    }
+}
```
matklad, line 147: Not a huge fan of these impls: they are public API, but we only want them as impl details. Specific problems with `base64::DecodeError`:

* the user might get an `base64::DecodeError` in a completely different part of their application, and they would be able to convert it to `workspaces::Error`. That is, this impl says that any base64 error is necessary a workspaces error, but that's not the case
* we might migrate of `base64` crate to some other impl of base64. We'd have to change public API for that.

workspaces/src/error/impls.rs
```diff
@@ -0,0 +1,199 @@
...
+            ErrorRepr::Custom { error, .. } => Some(error.as_mut()),
+            ErrorRepr::Full { error, .. } => Some(error.as_mut()),
+            _ => None,
+        }
+    }
+
+    /// Returns a mutable reference to the inner error (if any) downcasted to the type provided
+    pub fn downcast_mut<T: std::error::Error + Send + Sync + 'static>(&mut self) -> Option<&mut T> {
+        self.get_mut()?.downcast_mut()
+    }
```
matklad, line 128: Do we need these in the first version? I'd probably stick to just proving `std::error::Error` and add further downcasting in follow ups, if at all necessary. 

I do think these are useful, and should be added eventually, but they also seem not really necessary for the first cut

workspaces/src/network/betanet.rs
```diff
@@ -21,7 +21,7 @@ pub struct Betanet {
 }
 
 impl Betanet {
-    pub(crate) async fn new() -> anyhow::Result<Self> {
+    pub(crate) async fn new() -> crate::result::Result<Self> {
```
matklad, line 24:
```suggestion
    pub(crate) async fn new() -> crate::Result<Self> {
```

I think we  should re-export result/error at root, so that call-site could look like `workspaces::Result`

workspaces/src/error/mod.rs
```diff
@@ -0,0 +1,89 @@
+//! All errors that can occur within workspaces, including but not limited to
+//! the following: IO, RPC, Execution, Sandbox, DataConversion errors.
+
+mod impls;
+
+use std::borrow::Cow;
+
+/// A list specifying general categories of NEAR workspace error.
+#[derive(Clone, Debug, PartialEq, thiserror::Error)]
```
matklad, line 9: Feels a bit weird to derive PartialEq, but not Eq. I think it's OK to commit to Eq here.

workspaces/src/operations.rs
matklad: That's how reqwest works: 

https://docs.rs/reqwest/latest/src/reqwest/async_impl/request.rs.html#441-457

This seems like a nice API, from the caller's perspective -- that extra `?` would be useless like 99.999% of the time, and you'd have to type it for every `.json`. 

The builder here is kinda equivalent to calling a single function with a whole bunch of paramters. As you'd have a single `?` when calling the fn, there's no need to pessimize the builder. 

We do need to make sure that we use this pattern of infailable builder consistently .

matklad (approved):
LGTM! This now feels like something we can easily support&extend in the future, thanks for working on this!

workspaces/src/error/mod.rs
austinabell: I suppose my comment could be targeted more specifically at the issue that `CallExecutionDetails` execution errors get automatically converted into the `workspaces::Result`, which is why I indicate this as a motivation. For example, this test passes:

```rust
#[test(tokio::test)]
async fn test_failure() -> anyhow::Result<()> {
    let account = worker.dev_create_account().await?;

    let res = account.call(&worker, account.id(), "doesn't_exist").transact().await;
    assert!(matches!(res.unwrap_err().kind(), workspaces::error::ErrorKind::Execution));

    Ok(())
}
```

and it gives developers no way of inspecting the error message other than going through formatting the error into a string somehow. I generally think the API holistically needs to be more consistent and streamlined because there are a few inconsistencies that will make it harder for devs to use.

workspaces/src/operations.rs
austinabell: okay, I'll concede this, I suppose if keeping this pattern, introducing a new generic for the JSON type probably isn't ideal.

There is one thing that can change which is changing the inner type of `Function` to be a `Result` instead of keeping redundant data and performing operations when the serialization has already failed. This is optional but is what `request` does in the link you provided also. Might make things more maintainable in the future?

workspaces/src/error/mod.rs
matklad: Hm, that shoudn't automatically call `try_into_result`, that should be done by the user: 

```rust
let res = account.call(&worker, account.id(), "doesn't_exist").transact()
  .await? // this can fail if the sandbox is down
  .success()? // this will fail if the tx got to the network, but was faulty
;
```

`CallExecutionDetails` should be `#[must_use]`, and it shouldn't have data-erasign API like

```
fn try_into_result(self) -> Result<Self>
```

good catch!

austinabell:
https://github.com/near/workspaces-rs/pull/149#discussion_r936825271 is the only blocking comment from me, everything else is good enough to come in. Can also pull in and tackle in a separate PR since that is an issue that was not technically introduced in this PR, and these changes don't make it any worse

workspaces/src/error/mod.rs
matklad: It's not only the question of error message: the user should be able to tell, eg, how much gas was spend by a failed transaction.

ChaoticTempest: https://github.com/near/workspaces-rs/pull/150 was gonna resolve the `#[must_use]` so I'll leave it for that after I sync it up with all the recent changes from this current PR

austinabell (approved):

Issue #138, opened 2022-05-11
Hard to pull error message from `CallExecutionDetails`


austinabell:
There are methods to view each failure outcome, but even then it's janky to get to a result that can be used. `ExecutionOutcome` is not exactly easy to pull information needed.

Likely goes in tandem with #100 since the types for a lot of these can be avoided to be obfuscated as they are, because it's unclear to distinguish between result types

Also related to #131 since the error from execution is from the result, so the pattern is just that you have to format that error and not be able to distinguish where the error came from

Some of these issues can be seen from [near/near-sdk-rs@`3490844` (#789)](https://github.com/near/near-sdk-rs/pull/789/commits/3490844b2715e059a6a050cfb32cb21a988480eb) commit

Issue #131, opened 2022-05-10
Unintuitive API for call results


willemneal:
As noted by this diff:

[https://github.com/AhaLabs/bootme/pull/1/commits/6e970407d44a42b04005430c806107fa6c4cc207](https://github.com/AhaLabs/bootme/pull/1/commits/6e970407d44a42b04005430c806107fa6c4cc207)

I had expected to execute the call and then check if its call result failed. Whereas instead `transact` returned an Error.

Does this mean that `res.is_failure()` can never be true because the call result won't arrive due to the error being returned? Shouldn't user's contract errors be treated as normal expected behavior?

Something that would make this more clear is `transact_raw` in order to get back the raw result from RPC, which already seems to be the case if the transaction succeeds.

ChaoticTempest:
> I had expected to execute the call and then check if its call result failed. Whereas instead transact returned an Error.
> 
>Does this mean that res.is_failure() can never be true because the call result won't arrive due to the error being returned? Shouldn't user's contract errors be treated as normal expected behavior?

This is true, I'm going through the UX of how results are handled from calling into a transaction, and how its being done right now might be problematic for what you're mentioning. Originally, was going to add all the details into the error object included in the Result, but with https://github.com/near/workspaces-rs/issues/100, think its worth going back the other way and force users to call into a separate method to alert that the transaction succeeded.

So something like this should be preferable:
```
#[must_use]
struct CalExecutionDetails { ... }
impl CalExecutionDetails  {
    fn success() -> anyhow::Result<...>;
}
```
Only downside is that this is a breaking change

> Something that would make this more clear is transact_raw in order to get back the raw result from RPC, which already seems to be the case if the transaction succeeds.

could be an alternative to the breaking API, but would add `transact_raw` to everywhere that also has `transact`

PR #150, opened 2022-06-15
feat: must use CallExecution* type

ChaoticTempest:
This is not necessarily an API-breaking change but semantically would change how errors are being bubbled up. This adds the `#[must_use]` to `CallExecution*` types so that the users are forced to handle it. The previous way of turning the `CallExecution*` type into a `Result` was not so great since we'd end up consuming the `CallExecution*` info along the way if a user wanted to handle the error but also have the details of the execution.

Added `ok()` and `assert_ok()` methods to use the `CallExection*` objects if the user does not care about the results contained within them. They'll still have access to the execution info regardless


ChaoticTempest:
This should also resolve https://github.com/near/workspaces-rs/issues/138 and https://github.com/near/workspaces-rs/issues/131

workspaces/src/worker/impls.rs
```diff
@@ -81,8 +81,8 @@ where
                 deposit.unwrap_or(DEFAULT_CALL_DEPOSIT),
             )
             .await
+            .map(CallExecutionDetails::from)
             .map_err(crate::error::Error::from)
-            .and_then(CallExecutionDetails::from_outcome)
     }
```
itegulov, line 87: Any reason why you decided to switch the calls? The ordering shouldn't matter, right?

workspaces/tests/optional_args.rs
```diff
@@ -31,8 +32,17 @@ async fn test_empty_args_error() -> anyhow::Result<()> {
...
+
+    match res.ok() {
+        Ok(()) => panic!("Expected error: Failed to deserialize input from JSON"),
+        Err(err) => match err {
+            workspaces::error::Error::ExecutionError(msg) => {
+                assert!(msg.contains("Failed to deserialize input from JSON"));
+            }
+            other => panic!("Expected ExecutionError, got: {:?}", other),
+        },
+    }
```
itegulov, line 45: This feels like a very awkward pattern. Why not add a corresponding `err()` function as well?

workspaces/src/result.rs
```diff
@@ -105,15 +119,18 @@ impl CallExecutionDetails {
         }
     }
 
-    /// Convert the execution details into a Result if its status is not a successful one.
-    /// Useful for checking if the call was successful and forwarding the error upwards.
-    fn try_into_result(self) -> Result<Self> {
+    /// Checks whether the transaction was successful, if not returns a Result with the
+    /// corresponding error that occured.
+    pub fn ok(&self) -> Result<()> {
```
austinabell, line 124: This is a very strange pattern, the goal is to require the developer to use these `ok()` functions to check for errors and retrieve the inner values through another call? This feels very janky

workspaces/src/result.rs
```diff
@@ -105,15 +119,18 @@ impl CallExecutionDetails {
...
     }
 
-    pub(crate) fn from_outcome(outcome: FinalExecutionOutcomeView) -> crate::result::Result<Self> {
-        Self::from(outcome).try_into_result()
+    /// Assert that the transaction was successful, and that everything is good to go.
+    pub fn assert_ok(&self) {
+        if let Err(err) = self.ok() {
+            panic!("Transaction was not successful due to: {:?}", err);
+        }
     }
```
austinabell, line 134:
```suggestion
```
same as other

workspaces/src/worker/impls.rs
ChaoticTempest: Yeah the ordering doesn't matter here, this was more to switch over to the new style of error handling where if a user doesn't care about the error or the transaction details itself, they can just use `ok()`. Before, this was being forced handled via `from_outcome`, but that made the transaction details not available. Now, with `ok()`, the details are available and the user is "forced" to handle it

workspaces/src/result.rs
```diff
@@ -39,17 +41,29 @@ impl<T> CallExecution<T> {
     pub fn is_failure(&self) -> bool {
         self.details.is_failure()
     }
+
+    /// Checks whether the transaction was successful, if not returns a Result with the
+    /// corresponding error that occured.
+    pub fn ok(&self) -> Result<()> {
+        self.details.ok()
+    }
```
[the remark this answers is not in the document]
ChaoticTempest: Yeah this might not have been the best way of going about this one. I just added it because `CallExecution` was similar to `CallExecutionDetails`, but then again, we already have the `into_result` function here which does pretty much the same thing. I'll just get rid of this then

workspaces/src/result.rs
ChaoticTempest: The intent of this call was for them to call into `ok()` so that they don't trigger the `#[must_use]` warning while ignoring the rest of the execution details. Now, if they want to use the execution details, they would normally just reference all the other details instead of using `ok()`. Like if they just want to grab the details of the execution:
```rust
let details = contract.call(...)
   .transact()
   .await?;

println!("gas burnt: {}", details.total_gas_burnt);
```
or if they don't care at all about the details, but just want to know if the execution succeeded or not:
```rust
contract.call(...)
   .transact()
   .await?
   .ok()?;
```
It's just one way or the other, but not both at the same time, since we already have functions like `json`/`borsh` that allows us to deal with the error similarly if that makes sense.

Might not be the best way of expressing that though.

ChaoticTempest:
@itegulov @austinabell Let me know what you guys think of this now. `CallExecutionDetails` just has `executed()` so people can forward the errors up for convenience, `ok()`/`err()` to explicitly handle the errors with the details included

workspaces/src/result.rs
```diff
@@ -100,15 +103,34 @@ impl CallExecutionDetails {
...
-    fn try_into_result(self) -> Result<Self> {
+    /// Checks whether the transaction was executed successfully. This will return
+    /// a `Ok(())` if the transaction was successful, and a `Err(Error)` if it was not.
+    ///
+    /// This acts a convenience function to allow us to forward errors when we do not
+    /// care about the details of the transaction at all.
+    pub fn executed(&self) -> Result<()> {
         self.try_into_success_value()?;
-        Ok(self)
+        Ok(())
```
austinabell, line 113: It's the same error type, isn't it?

can just do (can't add suggestion on these lines):
```rust
self.try_into_success_value()
```

workspaces/src/result.rs
```diff
@@ -100,15 +103,34 @@ impl CallExecutionDetails {
...
+    }
+
+    /// Checks whether the transaction execution failed. This will return
+    /// `Some((CallExeuctionDetails, Error))` if it did fail, and `None` if it succeeded.
+    pub fn err(self) -> Option<(Self, Error)> {
+        match self.try_into_success_value() {
+            Err(err) => Some((self, err)),
+            Ok(_) => None,
+        }
     }
```
austinabell, line 134: I'm very skeptical about the naming and APIs for these. These do not seem intuitive to me.

So it seems like as a developer, you use `json`, `borsh`, or `raw_bytes` methods to get a result but `executed` if you don't care about the result data. Intuitively feels like this should be named more descriptively as `into_result` or something, where maybe we can return some of the data about the successful outcome that would otherwise be dropped.

As for the `ok` and `err` functions, I don't have any clear idea of how these would be used, and it's very confusing that they have different return types. This will likely be confusing with the [err](https://doc.rust-lang.org/std/result/enum.Result.html#method.err) and [ok](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) methods on the Rust `Result` type.

ChaoticTempest: definitely the naming could be improved. I feel like `executed` could stay since that conveys that we're expecting to have executed correctly and we include something like `into_result -> Result<SuccessfulExecutionOutcome>` if people want to make use of all the details. As for `ok`/`err`, not sure what's a better name. Maybe `handle`/`err` would be better where we can still keep `err` since that just returns Option like `std::result::Result` does

workspaces/src/result.rs
ChaoticTempest: This is actually returning `Result<&str>` so wouldn't be able to forward it

workspaces/src/result.rs
ChaoticTempest: Alright, instead of returning `(ExecutionDetails, Error)`, the details is now in the error object via `Error::detailed` and users would be able to grab it via `error.details()`. Got rid of `ok` as well since that didn't make sense and added `into_result() -> Result<ExecutionDetails>` instead. This should make for cleaner API. Let me know what you think

workspaces/src/result.rs
```diff
@@ -100,15 +105,31 @@ impl CallExecutionDetails {
...
+    pub fn into_result(self) -> Result<Self> {
         self.try_into_success_value()?;
         Ok(self)
     }
 
-    pub(crate) fn from_outcome(outcome: FinalExecutionOutcomeView) -> Result<Self> {
-        Self::from(outcome).try_into_result()
+    /// Checks whether the transaction execution failed. This will return
+    /// `Some((CallExeuctionDetails, Error))` if it did fail, and `None` if it succeeded.
+    pub fn err(self) -> Option<Error> {
```
austinabell, line 128: This comment is out of date. I would assume that this API would return something like `Option<CallExecutionDetails>` so perhaps this is an indication that the API is unintuitive. The fact that someone would have to check this error, match that the error kind is an execution error, then get details of the error (which is implicit from the previous step) seems like a huge leap for a developer.

austinabell (commented):
Going to open another issue because the types that are forwarded from the RPC types seem very unintuitive to use. So unclear how to proceed, for example, we are awaiting transactions to be fully completed, but we still have the intermediate execution statuses nested within the `CallExecutionDetails` structure.

ChaoticTempest:
New changes based on https://github.com/near/workspaces-rs/issues/183. Also took the liberty of renaming `CallExecution*` to be just `Execution*` since they're not purely just execution details from a call.

So after this, we have `ExecutionFinalResult` to allow us to do `into_result() -> Result<ExecutionSuccess, ExecutionFailure>`. Also note that I had `ExecutionFinalResult` could have just been a type alias of `ExecutionFinalResult` like the `Success` and `Failure` types, but `#[must_use]` would end up being applied to success/failure types as well which wasn't ideal.

workspaces/src/result.rs
```diff
@@ -149,6 +106,159 @@ impl CallExecutionDetails {
...
+    }
+}
+
+impl<T> Deref for ExecutionResult<T> {
+    type Target = ExecutionDetails;
+
+    fn deref(&self) -> &Self::Target {
+        &self.details
+    }
+}
```
austinabell, line 152: This feels like a strange impl to have, as it would confuse the API as these are very different types. Why not just expose the functions from `ExecutionResult` that you want from `ExecutionDetails`?

workspaces/src/error/impls.rs
```diff
@@ -64,24 +79,35 @@ impl Error {
...
     }
 
+    /// Get the associated execution details of this error. Usually found with
+    /// an error occuring from executing a transaction.
+    pub fn details(&self) -> Option<&ExecutionFailure> {
+        match &self.repr {
+            ErrorRepr::Detailed { error, .. } => Some(error),
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
ChaoticTempest: oops, originally had some of the return types returning/wrapping the Failure value into Error, but seems like I fully got rid of it, and this was no longer necessary

PR #152, opened 2022-06-20
release: 0.3.1

ChaoticTempest:
Gonna do a patch release to include the fix for rpc timeouts before the error handling https://github.com/near/workspaces-rs/pull/149 gets pulled in since that one has a decent amount of breaking changes and best if users from `0.3.0` could make use of the fixes since.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## [Unreleased]
 
+## [0.3.1] - 2022-06-20
+
+### Added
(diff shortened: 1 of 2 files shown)


austinabell (approved):
Unsure if the additions should be considered a patch release, but I can't really think ahead as to why this would be an issue in this case myself, can you?

ChaoticTempest:
yeah, it shouldn't be an issue since the changes are just additive and aren't a backwards compatibility issue, so we should be good to go

PR #154, opened 2022-06-21
fix: avoid creating new client for every request & hyper issues

ChaoticTempest:
basing this one off of https://github.com/near/workspaces-rs/pull/105 since we never got to pull it in. This also includes a fix for hyper related: `dispatch dropped without returning error`. The git diff looks horrendous trying to merge into the original PR due to rebasing/squashing, so this one would be faster to just push in


austinabell (approved):
utACK

PR #169, opened 2022-07-18
chore: Update Deps Along w/ M1 Support

ChaoticTempest:
This updates nearcore deps to the latest, and also sandbox dep where we now support M1.

Also, since updating to the latest nearcore, there seems to be a bug where the `EnvFilter` in nearcore ignores the case for `stats` target but works for others. So this is causing some stats to be logged every second. Minor annoyance, but will see what's causing it and fix it in the future on a patch release. Getting out M1 support for now in the newest minor patch is more important

### DIFF
--- a/workspaces/build.rs
+++ b/workspaces/build.rs
@@ -3,7 +3,6 @@ fn main() {
     let env_bin = std::env::var("NEAR_SANDBOX_BIN_PATH").is_ok();
     if !doc_build && !env_bin && cfg!(feature = "install") {
         // TODO Update commit to stable version once binaries are published correctly
-        near_sandbox_utils::install_with_version("master/97c0410de519ecaca369aaee26f0ca5eb9e7de06")
-            .expect("Could not install sandbox");
+        near_sandbox_utils::install().expect("Could not install sandbox");
     }
 }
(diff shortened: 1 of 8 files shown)


workspaces/build.rs
```diff
@@ -1,8 +1,8 @@
 fn main() {
     let doc_build = cfg!(doc) || std::env::var("DOCS_RS").is_ok();
-    if !doc_build && cfg!(feature = "install") {
+    let env_bin = std::env::var("NEAR_SANDBOX_BIN_PATH").is_ok();
```
austinabell, line 3: why is this additional check needed?

austinabell (approved):

workspaces/build.rs
ChaoticTempest: just thought it'd be nice to have if a user specified the env var so as to not trigger an install if not necessary, but I probably should add the if-env-var changed, a rebuild should occur. Would you see any potential footguns with this?

austinabell: I thought this was already checked? I guess it doesn't hurt to have redundancy to make sure it's short-circuited

PR #170, opened 2022-07-20
release: 0.4.0

ChaoticTempest:
Will do a release after this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 - internally no longer creating a new RPC client per call: https://github.com/near/workspaces-rs/pull/154
 - upped near dependencies to fix transitive vulnerabilities: https://github.com/near/workspaces-rs/pull/169
 
+### Changed
+- Default sandbox version is now using commit hash master/13a66dda709a4148f6395636914dca2a55df1390 (July 18, 2022): https://github.com/near/workspaces-rs/pull/169
+
 ## [0.3.1] - 2022-06-20
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [Unreleased]
 
+## [0.4.0] - 2022-07-20
+
+### Added
+- Mac M1 Support: https://github.com/near/workspaces-rs/pull/169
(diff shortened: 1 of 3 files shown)


ChaoticTempest:
hmm, after recent additions, the macOS tests seems to be pretty non-deterministic these days... I'll have to test this later, but it does seem like the instance of macOS GH has is really slow since it takes hours to publish the binaries in nearcore

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
 
+## [0.4.0] - 2022-07-20
+
+### Added
+- Mac M1 Support: https://github.com/near/workspaces-rs/pull/169
+- Added `Account::secret_key` to grab the account's secret key: https://github.com/near/workspaces-rs/pull/144
+- `Debug`/`Clone` impls for `Account`/`Contract`, and `Debug` for `Worker`: https://github.com/near/workspaces-rs/pull/167
+- `ExecutionOutcome::tokens_burnt` is now available: https://github.com/near/workspaces-rs/pull/168
+
+### Fixed
```
austinabell, line 13: Should we also indicate that the default sandbox version has changed?

austinabell (approved):
Release changes LGTM. Definitely worth looking into if the recent changes have caused the increase in non-deterministic issues so we don't push that to developers

austinabell (changes requested):
having an issue compiling with near-crypto 0.14 on ubuntu:
```
error[E0599]: no function or associated item named `from_bytes` found for struct `ed25519_dalek::Signature` in the current scope
  --> /home/austinabell/.cargo/registry/src/github.com-1ecc6299db9ec823/near-crypto-0.14.0/src/test_utils.rs:59:62
   |
59 |                 Signature::ED25519(ed25519_dalek::Signature::from_bytes(&SIG).unwrap())
   |                                                              ^^^^^^^^^^ function or associated item not found in `ed25519_dalek::Signature`
```

looking through docs the method doesn't exist, so not sure where this is referenced. Potentially missing a feature?

austinabell (approved):
nvm seems to be some dep mismatch, removing lockfile cleared

ChaoticTempest:
disabling the macos-latest test target since we can resolve that later

Issue #163, opened 2022-07-09
No way to manually override nonce for transactions


austinabell:
When sending transactions in parallel, it's very important that nonces can be overridden or else they will invalidate each other. Currently, the logic is handled like this:

[https://github.com/near/workspaces-rs/blob/c14fe2aa6cdf586028b2993c6a28240f78484d3e/workspaces/src/rpc/client.rs#L461-L476](https://github.com/near/workspaces-rs/blob/c14fe2aa6cdf586028b2993c6a28240f78484d3e/workspaces/src/rpc/client.rs#L461-L476)

This means that the nonce is pulled from the query and if two transactions query at the same time before sending the tx with `nonce + 1`, the second one to submit the transaction will be invalid.

It would be ideal if there was some mechanism to be able to send txs without waiting on the result, and have it available what the next nonce is. Not sure if this is the devX we want, but it would be ideal if we could parallelize sending the transactions without just exposing the option to set a nonce explicitly and require developers to understand how the nonce mechanics work.

PR #173, opened 2022-08-05
fix: Parallel tests should not wait on each other

ChaoticTempest:
This will resolve https://github.com/near/workspaces-rs/issues/163 without requiring the user to specify the nonce manually. The nonce is a value we can cache locally and only need to be retrieved once, incrementing it from thence onwards.

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -1,9 +1,8 @@
-use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::time::Duration;
 
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
@@ -33,15 +32,12 @@ use crate::types::{AccountId, InMemorySigner, Nonce, PublicKey};
 pub(crate) const DEFAULT_CALL_FN_GAS: Gas = 10_000_000_000_000;
 pub(crate) const DEFAULT_CALL_DEPOSIT: Balance = 0;
 
-lazy_static::lazy_static! {
-    static ref ACCESS_KEYS: Mutex<HashMap<AccountId, Nonce>> = Mutex::new(HashMap::new());
-}
-
 /// A client that wraps around [`JsonRpcClient`], and provides more capabilities such
 /// as retry w/ exponential backoff and utility functions for sending transactions.
 pub struct Client {
     rpc_addr: String,
     rpc_client: JsonRpcClient,
+    access_key_nonces: RwLock<HashMap<AccountId, Mutex<Nonce>>>,
 }
 
 impl Client {
@@ -52,6 +48,7 @@ impl Client {
         Self {
             rpc_client,
             rpc_addr: rpc_addr.into(),
+            access_key_nonces: RwLock::new(HashMap::new()),
         }
     }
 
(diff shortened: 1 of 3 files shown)


workspaces/src/rpc/client.rs
```diff
@@ -418,6 +422,35 @@ pub(crate) async fn access_key(
...
+
+        // Fetch latest block_hash since the previous one is now invalid for new transactions:
+        let block = client.view_block(Some(Finality::Final.into())).await?;
+        let block_hash = block.header.hash;
+        return Ok((block_hash, *nonce));
+    }
+    drop(nonces);
+
+    let (access_key, block_hash) = access_key(client, account_id.clone(), public_key).await?;
+    let mut nonces = client.access_key_nonces.write().await;
```
austinabell, line 446: So since all futures queue up to read first and all miss this if let statement, all txs after the first will fail due to an invalid nonce, logs if you print out the error:
```rust
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
TX err: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 6000001, ak_nonce: 6000001 }]
```

There is a slightly larger issue with the logic of workspaces right now. This error is not temporal and should error when it's hit the first time. Because the async function that is retried will re-enter this block and hit the if statement the second time around, this bug is hidden.

I've fixed this and switched the inner type from a `Mutex<Nonce>` to `AtomicU64` to optimize the incrementing.

diff:
```
diff --git a/workspaces/src/rpc/client.rs b/workspaces/src/rpc/client.rs
index 84f9909..933c447 100644
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -1,8 +1,9 @@
 use std::collections::HashMap;
 use std::fmt::Debug;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::RwLock;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
@@ -39,7 +40,7 @@ pub struct Client {
     rpc_addr: String,
     rpc_client: JsonRpcClient,
     /// AccessKey nonces to reference when sending transactions.
-    access_key_nonces: RwLock<HashMap<AccountId, Mutex<Nonce>>>,
+    access_key_nonces: RwLock<HashMap<AccountId, AtomicU64>>,
 }
 
 impl Client {
@@ -430,25 +431,20 @@ async fn fetch_tx_nonce(
     account_id: AccountId,
     public_key: near_crypto::PublicKey,
 ) -> anyhow::Result<(CryptoHash, Nonce)> {
-    let nonces = client.access_key_nonces.read().await;
+    let mut nonces = client.access_key_nonces.write().await;
     if let Some(nonce) = nonces.get(&account_id) {
-        let mut nonce = nonce.lock().await;
-        *nonce += 1;
+        let nonce = nonce.fetch_add(1, Ordering::SeqCst);
 
         // Fetch latest block_hash since the previous one is now invalid for new transactions:
         let block = client.view_block(Some(Finality::Final.into())).await?;
         let block_hash = block.header.hash;
-        return Ok((block_hash, *nonce));
-    }
-    drop(nonces);
+        Ok((block_hash, nonce + 1))
+    } else {
+        let (access_key, block_hash) = access_key(client, account_id.clone(), public_key).await?;
+        nonces.insert(account_id, AtomicU64::new(access_key.nonce + 1));
 
-    let (access_key, block_hash) = access_key(client, account_id.clone(), public_key).await?;
-    let mut nonces = client.access_key_nonces.write().await;
-    nonces
-        .entry(account_id)
-        .or_insert_with(|| Mutex::new(access_key.nonce + 1));
-
-    Ok((block_hash, access_key.nonce + 1))
+        Ok((block_hash, access_key.nonce + 1))
+    }
 }
 
 pub(crate) async fn retry<R, E, T, F>(task: F) -> T::Output
@@ -472,7 +468,7 @@ pub(crate) async fn send_tx(
             signed_transaction: tx,
         })
         .await
-        .map_err(|e| anyhow::anyhow!(e))
+    .map_err(|e| anyhow::anyhow!(e))
 }
 
 pub(crate) async fn send_tx_and_retry<T, F>(

```

This optimizes the test case by not having to wait on the failed txs to finalize and going through the extra mutex, which cuts the execution time a few seconds.

We could optimize this a little more by switching the data structure to some concurrent hash map to optimize sending from different accounts, but that might be a bit premature

ChaoticTempest: Oh nice catch, but there's one thing I corrected in the new commit since yours requires a `write` lock each fetch, which there could be contention with when there's multiple queries into different `account_id`'s. There's a different fix to this where we handle the case when multiple writers try to get the write lock the first time: just send them to the next loop. Let me know what you think. And yeah, I didn't go with a concurrent hash map because of the premature optimization

ChaoticTempest:
Talked with @volovyks about ideal designs with how the nonce is handled for multi transactions. This ended up becoming the most ideal solution where we cache it, then invalidate it if the nonce was invalid. NAJ already has the same behavior so  workspaces-js has it as well as a consequence. Probably need to revisit this in the future when we add something like `transact_async` but for now, this is a general improvement where we less contention when utilizing the same access keys to send transactions.

@austinabell let me know if you have any concerns here but otherwise would be nice if we got this in. Only PR left before 0.5.0 release

austinabell (approved):

PR #176, opened 2022-08-11
Added a reminder to rebuild before running tests

ryancwalsh:


README.md
```diff
@@ -32,6 +32,12 @@ rustup default stable-x86_64-apple-darwin
 ### WASM compilation not supported
 `workspaces-rs`, the library itself, does not currently compile to WASM. Best to put this dependency in `[dev-dependencies]` section of `Cargo.toml` if we were trying to run this library alongside something that already does compile to WASM, such as `near-sdk-rs`.
 
+## Important reminder (a "gotcha")
+
+Running a Workspaces-rs test (such as via `RUST_BACKTRACE=full cargo test --package your_package --test integration-tests -- some_specific_test --exact --show-output` or by clicking "Run Test" in VSC) does _not_ compile your code first. 
+
+Always [rebuild](https://www.near-sdk.io/building/basic-build) before running your tests, or else you'll be testing against stale code.
```
ChaoticTempest, line 39: I think you're talking about how contracts need to be re-compiled if you change them? - workspaces/test code will automatically get compiled if they are changed by the compiler. If you are talking about contract code, this is pretty inherent in how you have to point to the pre-compiled WASM file in the test code, so we don't need this blurb to describe this here

ryancwalsh: My impression is that we (the NEAR ecosystem at large) are trying to attract many more developers to Rust (including me, who just started in Rust a number of weeks ago). So my idea here is to smooth out the experience and try to reduce the gotchas. In certain ways, Rust is a complete pain to code in compared to other languages, and I found myself multiple times wondering why a test was failing and then realizing that I hadn't recompiled the contract. It could be that I'm an idiot and nobody else will be, but my suggestion here was just to assume that making everything as foolproof as possible will help the ecosystem the most.

An alternative to having a note in the Readme might be to offer tooling that we recommend to use that will always ensure that tests are run against fresh Wasm instead of stale code.

ChaoticTempest: gotcha, maybe in that case we can just have a note saying:
> `deploy`/`dev_deploy` functions takes a path to a WASM contract file which does not recompile our contract code if it changes. If we want to listen to contract code changes, use something like `worker.compile_project("/path/to/rs/project").await?;`

just so we can be explicit about what needs to be recompiled here. There's things you can do to make the contract be recompiled, like watching a file for changes to run cargo build but it would be bad devX to recommend that.

cc @austinabell since I remember you had some opinions about including a section about re-compilation in the README

ChaoticTempest: Also, this is not just an inherent problem to the rust version of workspaces. This is required as well with workspaces-js that uses either the `near-sdk-rs` or `near-sdk-js`

austinabell: yeah I don't think I have an opinion of what is included in the README about re-compilation, definitely your call!

This topic did get me thinking though if we should be more specific with the unstable `compile_project` API, since it assumes a Rust contract. Although using a JS contract with Rust tests is insanely unlikely, I wonder if there is some way we should future-proof. Probably not, just thinking out loud

ryancwalsh: @ChaoticTempest and @austinabell I'm still leaning towards what I originally suggested above (at https://github.com/near/workspaces-rs/pull/176/files). 

If you'd prefer something different, I'm happy to make edits, or if you'd prefer to abandon this PR and just make your own edits directly in code, that's fine with me. 

I think we're probably all in sync about this goal:

> My impression is that we (the NEAR ecosystem at large) are trying to attract many more developers to Rust (including me, who just started in Rust a number of weeks ago). So my idea here is to smooth out the experience and try to reduce the gotchas.

and I'm sure there are many different approaches that would be fine. :-)

ChaoticTempest:
closed with https://github.com/near/workspaces-rs/pull/186

PR #177, opened 2022-08-11
fix: tests erroring out on newest rust v1.63.0

ChaoticTempest:
This fixes the current builds failing with the release of 1.63.0 with newer clippy warnings.
- also bumped MSRV to 1.60 since near-primitives 0.14 requires it.


itegulov (approved):
LGTM. Was about to ask why our CI had not failed earlier, but then I saw that we never used MSRV in the pipeline.

austinabell (approved):

PR #180, opened 2022-08-16
release: 0.4.1

ChaoticTempest:
Will release 0.4.1 with the overflow error fix and without error handling included since that's a much bigger breaking change and should be in 0.5.0 instead

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,7 @@
 
 
 [Unreleased]: https://github.com/near/workspaces-rs/compare/0.4.1...HEAD
-[0.4.1]: https://github.com/near/workspaces-rs/compare/0.4.0-no-eh...0.4.1
+[0.4.1]: https://github.com/near/workspaces-rs/compare/0.4.0...0.4.1
 [0.4.0]: https://github.com/near/workspaces-rs/compare/0.3.1...0.4.0
 [0.3.1]: https://github.com/near/workspaces-rs/compare/0.3.0...0.3.1
 [0.3.0]: https://github.com/near/workspaces-rs/compare/0.2.1...0.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+## [0.4.1] - 2022-08-16
+
+### Added
+- Derive `Eq` on `AccountDetails` type: https://github.com/near/workspaces-rs/pull/177/files
+
+### Fixed
+- Fix macOS non-deterministic overflow error when starting up sandbox: https://github.com/near/workspaces-rs/pull/179
+
 ## [0.4.0] - 2022-07-20
 
 ### Added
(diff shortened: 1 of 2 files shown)


itegulov (approved):

CHANGELOG.md
```diff
@@ -81,7 +89,8 @@
 - Fix race condition when installing sandbox and running multiples tests at the same time. https://github.com/near/workspaces-rs/pull/46
 
 
-[Unreleased]: https://github.com/near/workspaces-rs/compare/0.4.0...HEAD
+[Unreleased]: https://github.com/near/workspaces-rs/compare/0.4.1...HEAD
+[0.4.1]: https://github.com/near/workspaces-rs/compare/0.4.0-no-eh...0.4.1
```
austinabell, line 93:
```suggestion
[0.4.1]: https://github.com/near/workspaces-rs/compare/0.4.0...0.4.1
```

typo?

austinabell (approved):

CHANGELOG.md
ChaoticTempest: this was to exclude the error handling PR. Not sure how to express that in the comparison besides starting from a commit hash that doesn't have it

austinabell: why would this need a different starting tag? doesn't 0.4.1 continue from 0.4.0? Or was 0.4 done manually to exclude a commit?

ChaoticTempest: yeah this is kinda funky. It just so happens the error handling commit is right after 0.4.0 commit, which means we can skip it and start from the commit after.

The actual crate release will just be cherry picked commits though, so the CHANGELOG is more of me trying to not include it within the reference frame of 0.4.0...0.4.1. The above is what I came up with by having a special tag at the start of after error handling just so we don't actually show it in the comparison link, but let me know if there's a better way of doing this

austinabell: Either way, if you've cherry-picked or removed that commit, wouldn't the history include 0.4.0 tag? I guess I'm just more curious about how this release will be created, since releasing this commit merged would include error handling?

ChaoticTempest: it's not going to be from the main branch but on a separate branch with all the cherry picked commits. Probably could just reference that branch instead for the comparison

austinabell: but what I'm saying is that you could compare the two tags, since the branch would have 0.4.0 in the commits? This isn't a blocking comment btw, feel free to do the release and merge if you're sure

ChaoticTempest: I think I mostly get what you're saying, so just to be clear you're saying we have something like:
`https://github.com/near/workspaces-rs/compare/0.4.0...0.4.1` where 0.4.1 is the tagged via the separate branch (with all the cherry-picked commits) right?

austinabell: Is that not what you were planning on doing anyway? I'm confused. Just do what you were planning on doing, was just trying to help avoid an unnecessary new tag

PR #181, opened 2022-08-19
feat: `worker` no longer required to be passed into arguments when transacting transactions

ChaoticTempest:
Also, no longer require importing `prelude`

### DIFF
--- a/workspaces/src/network/sandbox.rs
+++ b/workspaces/src/network/sandbox.rs
@@ -12,9 +12,8 @@ use crate::network::server::SandboxServer;
 use crate::network::Info;
 use crate::result::{CallExecution, Result};
 use crate::rpc::client::Client;
-use crate::rpc::patch::ImportContractTransaction;
 use crate::types::{AccountId, Balance, InMemorySigner, SecretKey};
-use crate::{Account, Contract, Network, Worker};
+use crate::{Account, Contract, Worker};
 
 // Constant taken from nearcore crate to avoid dependency
 pub(crate) const NEAR_BASE: Balance = 1_000_000_000_000_000_000_000_000;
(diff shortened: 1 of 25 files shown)


workspaces/src/network/sandbox.rs
```diff
@@ -87,29 +86,35 @@ impl AllowDevAccountCreation for Sandbox {}
...
 #[async_trait]
 impl TopLevelAccountCreator for Sandbox {
-    async fn create_tla(&self, id: AccountId, sk: SecretKey) -> Result<CallExecution<Account>> {
+    async fn create_tla(
+        &self,
+        worker: Worker<dyn Network>,
+        id: AccountId,
+        sk: SecretKey,
+    ) -> Result<CallExecution<Account>> {
         let root_signer = self.root_signer()?;
```
austinabell, line 95: There's no test for this or `create_tla_and_deploy`. Definitely can be done in a separate PR but it's a bit unclear how this would be used by a developer. Would you mind adding a test for this to make sure it's usable?

My intuition is that this won't be very usable because the developer would have to do the coercion themselves. Might require a bit of re-working of this TLAC trait. Am I missing something here?

ChaoticTempest: So, there's one new thing with this PR, Networks like Sandbox/Testnet impl TLAC but worker has it's own custom impl of it but without impl the TLAC trait. So it just has the signature of `create_tla(id: AccountId, sk: SecretKey)` instead. With this, the user never really uses the TLAC trait

austinabell: Can we remove it in the next minor version then? Does it have any use? If these traits are going to be removed in the future, makes sense to do sooner than later?

ChaoticTempest: the TLAC trait is still required since we need an interface to call into the `create_tla` function for each respective network, and provides the ability to add custom networks down the line. So I don't see this trait being removed, but the `DevAccountDeployer` trait didn't make sense to stay so that one was removed

austinabell (approved):

workspaces/tests/deploy_project.rs
```diff
@@ -1,23 +1,22 @@
...
 #[test(tokio::test)]
 async fn test_dev_deploy_project() -> anyhow::Result<()> {
     let worker = workspaces::sandbox().await?;
     let wasm = workspaces::compile_project("./tests/test-contracts/status-message").await?;
     let contract = worker.dev_deploy(&wasm).await?;
 
-    let _res = contract
-        .call(&worker, "set_status")
+    contract
+        .call("set_status")
```
austinabell, line 12: Curious why this isn't required with `must_use`. Isn't the point of this PR to disallow not using the result?

PR #185, opened 2022-08-26
chore: Added convenience functions `Account::{from_secret_key, set_secret_key}`

ChaoticTempest:
These seemed pretty useful to add since `Account::from_file` didn't make sense if you already had the `SecretKey` value in memory. `Account::set_secret_key` is useful for changing the signer of transactions to use a different one if users had multiple secret keys

### DIFF
--- a/workspaces/src/types/account.rs
+++ b/workspaces/src/types/account.rs
@@ -220,16 +220,16 @@ impl Contract {
         &self.account.id
     }
 
-    /// Casts the current [`Contract`] into an [`Account`] type. This does
-    /// nothing on chain/network, and is merely allowing `Account::*` functions
-    /// to be used from this `Contract`.
+    /// Treat this [`Contract`] object as an [`Account`] type. This does nothing
+    /// on chain/network, and is merely allowing `Account::*` functions to be
+    /// used from this `Contract`.
     pub fn as_account(&self) -> &Account {
         &self.account
     }
 
-    /// Casts the current [`Contract`] into an [`Account`] type. This does
-    /// nothing on chain/network, and is merely allowing `Account::*` functions
-    /// to be used from this `Contract`.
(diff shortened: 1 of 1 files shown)


workspaces/src/types/account.rs
```diff
@@ -156,6 +169,12 @@ impl Account {
     pub fn secret_key(&self) -> &SecretKey {
         &self.signer.secret_key
     }
+
+    /// Sets the [`SecretKey`] of this account. Future transactions will be signed
+    /// using this newly provided key.
+    pub fn set_secret_key(&mut self, sk: &SecretKey) {
+        self.signer.secret_key = sk.clone();
```
austinabell, line 176:
```suggestion
    pub fn set_secret_key(&mut self, sk: SecretKey) {
        self.signer.secret_key = sk;
```

workspaces/src/types/account.rs
```diff
@@ -38,6 +38,19 @@ impl Account {
...
+    pub fn from_secret_key(
+        id: &AccountId,
+        sk: &SecretKey,
+        worker: &Worker<impl Network + 'static>,
+    ) -> Self {
+        Self {
+            id: id.clone(),
+            signer: InMemorySigner::from_secret_key(id.clone(), sk.clone()),
+            worker: worker.clone().coerce(),
+        }
```
austinabell, line 51: these should probably take owned values too, to avoid cloning if not necessary

austinabell (approved):

workspaces/src/types/account.rs
```diff
@@ -199,6 +227,13 @@ impl Contract {
         &self.account
     }
 
+    /// Casts the current [`Contract`] into an [`Account`] type. This does
```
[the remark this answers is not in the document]
ChaoticTempest: that's a good point. Reworded it, let me know what you think

PR #188, opened 2022-08-31
release: 0.5.0

ChaoticTempest:
This also bumps the sandbox version to 0.5.0 as well. Will make a release afterwards


ChaoticTempest:
Ideal if https://github.com/near/workspaces-rs/pull/173 also makes it in before this gets released

Issue #184, opened 2022-08-25
RPCaas support with API keys


austinabell:
Currently, our APIs for all networks are pointing at base free RPC nodes, where for more heavy-load or to connect with console, it would be beneficial to hit these endpoints.

API is opinionated on how this is done, but ideally this has little duplication because the only thing that changes for each network is the URL and adding API keys.

miraclx:
`near-jsonrpc-client` has `auth::ApiKey` that workspaces can build atop of. — https://github.com/near/near-jsonrpc-client-rs/blob/master/src/auth.rs

shariffdev:
This can be closed @frol

Issue #103, opened 2022-04-04
Can't create custom network


austinabell:
Types weren't exported and the API doesn't support creating. #102 moves towards this, but some things are still not accessible.

For example `NetworkClient` trait required to use things requires returning a `Client` reference, but there is no way to construct a Client externally.

ChaoticTempest:
Good point to mention. Some stuff might not be ready to allow users to construct their own network quite yet (such as the trait for top level account creation not having a good enough API), so was leaving this feature out for now. But definitely a good issue to tackle in the future to allow this

bharath-123:
@ChaoticTempest i think what @austinabell suggests is useful because we might want to run our own sandbox and persist state between integration test runs. what do you think?

ChaoticTempest:
This issue was more about creating a custom network where we can have custom behavior such as overriding calls into creating accounts via the helper contracts in testnet for instance. But for persisting state between sandbox, we have something like https://github.com/near/workspaces-rs/issues/64 which is currently blocked by how we can kill the sandbox after all tests have passed. You can use that if you need it now, but you'll potentially need to kill the sandbox yourself.

A simple way of getting the workflow you're suggesting (and having workspaces not manage the sandbox), is by specifying some configs in one of the initialization functions of workspaces. Probably an uncommon case to make use of, but could be worthwhile if you see value in wanting to start your own sandbox. Just note that this will end up making workspaces not have full control of sandbox, and then you will be the sole manager of it. This could mean resource leaks if you don't kill the sandbox after you're done with tests

PR #190, opened 2022-09-09
feat: Added API Keys

ChaoticTempest:
PR adds in `NEAR_RPC_API_KEY` to interact with RPC related services. Resolves https://github.com/near/workspaces-rs/issues/184

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -46,15 +46,20 @@ pub struct Client {
 }
 
 impl Client {
-    pub(crate) fn new(rpc_addr: &str) -> Self {
+    pub(crate) fn new(rpc_addr: &str) -> Result<Self> {
         let connector = JsonRpcClient::new_client();
-        let rpc_client = connector.connect(rpc_addr);
+        let mut rpc_client = connector.connect(rpc_addr);
+        if let Ok(api_key) = std::env::var("NEAR_RPC_API_KEY") {
+            let api_key = near_jsonrpc_client::auth::ApiKey::new(api_key)
+                .map_err(|e| ErrorKind::DataConversion.custom(e))?;
+            rpc_client = rpc_client.header(api_key);
+        }
 
-        Self {
+        Ok(Self {
             rpc_client,
             rpc_addr: rpc_addr.into(),
             access_key_nonces: RwLock::new(HashMap::new()),
-        }
+        })
     }
 
     pub(crate) async fn query_broadcast_tx(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- RPC API Keys used to interact with services such as Pagoda Console.
+
 ## [0.5.0]
 ### Added
 - Error handling with opaque `workspaces::error::Error` type: https://github.com/near/workspaces-rs/pull/149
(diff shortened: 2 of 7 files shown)


workspaces/src/rpc/client.rs
```diff
@@ -46,15 +46,20 @@ pub struct Client {
 }
 
 impl Client {
-    pub(crate) fn new(rpc_addr: &str) -> Self {
+    pub(crate) fn new(rpc_addr: &str) -> Result<Self> {
         let connector = JsonRpcClient::new_client();
-        let rpc_client = connector.connect(rpc_addr);
+        let mut rpc_client = connector.connect(rpc_addr);
+        if let Ok(api_key) = std::env::var("NEAR_RPC_API_KEY") {
```
miraclx, line 52: Using an env feels very much like a downstream feature, we should probably just accept API key input

```rust
let mut worker = workspaces::mainnet().await?;
worker.set_api_key(..)?;
```

Which then sets the header for the inner `NetworkClient`.

workspaces/src/rpc/client.rs
```diff
@@ -46,15 +46,20 @@ pub struct Client {
...
+            let api_key = near_jsonrpc_client::auth::ApiKey::new(api_key)
+                .map_err(|e| ErrorKind::DataConversion.custom(e))?;
```
austinabell, line 54: Since we don't have the ability to override the RPC URL, this feels like a half-baked solution that won't be usable by anyone. I wonder if we should have a separate endpoint, like `workspaces::mainnet`, except that it points to RPCaaS URL and requires an API key so this can't be misused

workspaces/src/rpc/client.rs
ChaoticTempest: This was mainly done for the security reason that workspaces would most likely be run in CI and could potentially expose these keys if people weren't careful about it. But then again, there might be cases where its easier to set this programmatically. I'll add it, but will need a warning not to use this in a public facing CI

workspaces/src/rpc/client.rs
ChaoticTempest: right, forgot that the Console will have it's own dedicated RPC URL instead of the defaults ones on `*.near.org`, so thought this would be low hanging fruit to add like this. This goes into the territory of custom networks with all of its quirks, but it does seem easy to just add something like `workspaces::custom` to point to a custom endpoint

austinabell: Yeah, or maybe providing reasoning for https://github.com/near/workspaces-rs/issues/103 so that this functionality could exist in a separate crate or any other community need. I feel like `custom` might not solve this issue since API keys are only applicable to the new RPCaaS endpoints so would be janky to include with that API.

Either way, if we do introduce this ability, it should be an `unstable` API as the API keys and endpoints might change in the future

workspaces/src/rpc/client.rs
miraclx: That makes sense, in that case the env var is sufficient. Was thinking of workspaces from the context of a client library, much like `near-api-js`, whereas it's much more than that and the target scope is mostly for testing, esp since, by default, it installs a sandbox. All good, then, feel free to skip over this comment.

frol:
Closing as it requires rebasing and addressing the review comments. Next attempt to resolve #184 should take all the above comments into account.

PR #192, opened 2022-09-13
feat: Contract from Account

joshuajbouw:
All this change does is simply allow for a `Contract` to be created from an `Account`.

This is beneficial for our needs as we would like to create root contracts in the sand box. I do not believe that this should cause any negative issues and simply opens up more creation abilities of `Contract`.

## Example

```
async fn make_evm_account() -> anyhow::Result<(Contract)> {
    let root: NearAccount = worker.root_account()?.into();
    
    let sk = SecretKey::from_random(KeyType::ED25519);
    let aurora_account: Contract = root
        .create_subaccount("aurora.root")
        .initial_balance(100_000_000_000_000_000_000_000_000)
        .keys(sk)
        .transact()
        .await?
        .into_results()?
        .into();
    Ok(aurora_account)
}
```

### DIFF
--- a/workspaces/src/types/account.rs
+++ b/workspaces/src/types/account.rs
@@ -30,7 +30,7 @@ impl fmt::Debug for Account {
 impl Account {
     /// Create a new account with the given path to the credentials JSON file
     pub fn from_file(
-        path: impl AsRef<std::path::Path>,
+        path: impl AsRef<Path>,
         worker: &Worker<impl Network + 'static>,
     ) -> Result<Self> {
         let signer = InMemorySigner::from_file(path.as_ref())?;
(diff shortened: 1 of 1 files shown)


workspaces/src/types/account.rs
```diff
@@ -30,7 +30,7 @@ impl fmt::Debug for Account {
 impl Account {
     /// Create a new account with the given path to the credentials JSON file
     pub fn from_file(
-        path: impl AsRef<std::path::Path>,
```
joshuajbouw, line 33: Note that this was redundant to have as it is already imported, so I had safely removed it.

joshuajbouw:
Going to close this as I had found that you can have the same result by doing:

```
async fn make_evm_account() -> anyhow::Result<(Contract)> {
    let wasm = std::fs::read(ENGINE_WASM_FILEPATH)?;
    let root: NearAccount = worker.root_account()?.into();
    
    let sk = SecretKey::from_random(KeyType::ED25519);
    let aurora_account: Account = root
        .create_subaccount("aurora.root")
        .initial_balance(100_000_000_000_000_000_000_000_000)
        .keys(sk)
        .transact()
        .await?
        .into_results()?;
    let aurora_contract = aurora_account.deploy(&wasm).await?.into_result()?;
    Ok(aurora_contract)
}
```

PR #202, opened 2022-09-22
feat: add view calls to an `Account`

joshuajbouw:
For whatever reason, you can not do a view call from an `Account`. This simply fixes that functionality.


ChaoticTempest (approved):
whoops, this might've slipped since `account.call().view()` was a thing

PR #203, opened 2022-09-23
feat: return `ExecutionFailure` error

joshuajbouw:
Right now, there is no way to be able to return the error, which will be useful for users. The best way to do this is to simply provide all the details that a transaction could fail through the `near_primitives::TxExecutionError`.

### DIFF
--- a/workspaces/src/error/mod.rs
+++ b/workspaces/src/error/mod.rs
@@ -4,6 +4,7 @@
 pub(crate) mod execution;
 mod impls;
 
+pub use near_primitives::errors::TxExecutionError;
 use std::borrow::Cow;
 
 use crate::result::ExecutionFailure;
--- a/workspaces/src/result.rs
+++ b/workspaces/src/result.rs
@@ -315,6 +315,13 @@ impl ExecutionSuccess {
     }
 }
 
+impl ExecutionFailure {
+    /// Returns the execution error as a [`near_primitives::errors::TxExecutionError`].
+    pub fn error(&self) -> &TxExecutionError {
+        &self.value
+    }
+}
+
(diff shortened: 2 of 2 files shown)


ChaoticTempest:
What is the use case for this? It's intended not to be exposed currently because we do not want to maintain this transitive dependency and potentially cause breaking changes if `near-primitives` decides to change down the line. So far, the only real use case I've seen is for debugging purposes where we can just `format!("{:?}", ExecutionFailure)` or forward it to the very top. Is there something special you need out of it?

joshuajbouw:
When testing the Aurora EVM, we need to know what errors are being returned. There are a plethora of errors that could possibly happen in a given execution, instead of testing if it simply just failed.

ChaoticTempest:
> When testing the Aurora EVM, we need to know what errors are being returned. A plethora of errors could happen in a given execution instead of testing if it simply failed.

hmm, but are you guys explicitly handling these errors case-by-case? I'm wondering why we'd need to do that in a test if that is the case, or are you guys using `workspaces-rs` for something other than tests?

> Specifically, we need the error that the contract returns. Might be a good idea to instead have a method that returns a contract error, if any?

Are you talking about the error returned from a contract's function that returns a `Result` and you're trying to grab the error variant? Then `TxExecutionError` is not what you're looking for. If your contract function returns a `Result` it can be retrieved the same way you retrieve other values that aren't `Result`s:
```rust
let outcome: Result<ContractValue, ContractError> = contract.call(...).transact().await?.json()?;
```

joshuajbouw:
Considering that we have been panic_utf8-ing on error and planning on fixing our contract API to return a result instead, yes, we could do that.

Issue #201, opened 2022-09-21
ExecutionOutcome ValueOrReceiptId::Value should have convenient methods for parsing outcome


ChaoticTempest:
Currently, `ExecutionOutcome` internal returned value is just being exposed purely as a `ValueOrReceiptId::Value(String)`, which does not offer all the convenience methods of `json`, `borsh`, or `raw_bytes` which makes it feel inconsistent with the other APIs. Ideally should add in a new `Value` type that offers these convenient methods.

PR #208, opened 2022-09-30
chore: Made ValueOrReceiptId::Value object consistent

ChaoticTempest:
This adds a new `Value` type with convenient APIs like `raw_bytes`, `borsh`, and `json` to the value of `ValueOrReceiptId::Value` variant for `ExecutionOutcome`.

Resolves https://github.com/near/workspaces-rs/issues/201

Note that this is a breaking API, so will require a new minor version for this PR

### DIFF
--- a/workspaces/src/result.rs
+++ b/workspaces/src/result.rs
@@ -17,7 +17,7 @@ pub type Result<T, E = crate::error::Error> = core::result::Result<T, E>;
 /// Execution related info as a result of performing a successful transaction
 /// execution on the network. This value can be converted into the returned
 /// value of the transaction via [`ExecutionSuccess::json`] or [`ExecutionSuccess::borsh`]
-pub type ExecutionSuccess = ExecutionResult<String>;
+pub type ExecutionSuccess = ExecutionResult<Value>;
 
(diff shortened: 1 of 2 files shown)


workspaces/src/result.rs
```diff
@@ -448,14 +447,57 @@ impl ExecutionOutcome {
...
+    /// Deserialize an instance of type `T` from bytes of JSON text sourced from the
+    /// execution result of this call. This conversion can fail if the structure of
+    /// the internal state does not meet up with [`serde::de::DeserializeOwned`]'s
+    /// requirements.
+    pub fn json<U: serde::de::DeserializeOwned>(&self) -> Result<U> {
+        let buf = self.raw_bytes()?;
+        serde_json::from_slice(&buf).map_err(|e| ErrorKind::DataConversion.custom(e))
+    }
+
+    /// Deserialize an instance of type `T` from bytes sourced from the execution
```
austinabell, line 481: Noting that `T` doesn't match the generic name `U`

austinabell (approved):

itegulov (approved):

PR #210, opened 2022-10-04
Chore: Upgrade to version 0.4.0 of near-jsonrpc-client, and version 0.15.0 for nearcore libraries

birchmd:
`near-jsonrpc-client` was officially released, so we do not need to depend on the `-beta.0` version any more.

This upgrade is important because cargo thinks that `0.4.0-beta.0` should be compatible with `0.4.0`, however they are not due to the different versions of the nearcore libraries. This means that `workspaces` fails to compile as a dependency of other crates (unless the `near-jsonrpc-client` version is fixed to the `beta` one in `Cargo.lock`).

Along with the upgrade, there were a few minor API changes in the nearcore libraries which meant we needed to make a few extra changes here (namely around encoding values as base64 / base 58).

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 anyhow = "1.0"
 borsh = "0.9"
 maplit = "1.0"
-near-units = "0.1.0"
+near-units = "0.2.0"
 serde = "1.0"
 serde_with = "1"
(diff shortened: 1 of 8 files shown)


ChaoticTempest (approved):
Thanks for upgrading these! The secret keys being changed will make this a breaking change, but nothing you have to worry about since this will be apart of a new minor release. Probably 0.6

PR #212, opened 2022-10-05
release: 0.5.1

ChaoticTempest:
Will cherry pick commits for 0.5.1 which only includes the fixes listed in the CHANGELOG


ChaoticTempest:
looks like we can't do a patch release for this since near-jsonrpc-client got updated to 0.4.0 which overrides 0.4.0-beta.0 when cargo tries to resolve deps

PR #218, opened 2022-10-11
feat: Add async builders for `view_*` functions

ChaoticTempest:
This adds the async builder `Query` which allows doing view calls into contract state, account, access keys, code, the view function call, and block. This uses `IntoFuture` to allow calling into `.await` at any point in the builder. Note, that `Query` builder takes a generic `<T>` for the method like `ViewState` since we can define custom builder functions per method type by doing `impl Query<ViewState> { ... }`

cc @matklad if you want to review my use of `IntoFuture` since you initially suggested it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,27 @@
 
 ## [Unreleased]
 
-## [0.5.0]
 ### Added
+
+- `view_*` builders have been added
+
+### Changed
+
+- `{Account, Contract, Worker}::view_state` moved `prefix` parameter into builder. i.e.
+
+```
+worker.view_state("account_id", Some(prefix)).await?;
+// is now
+worker.view_state("account_id")
+    .prefix(prefix)
+    .await?;
+// if prefix was `None`, then simply delete the None argument.
+```
+
+## [0.5.0]
+
+### Added
+
 - Error handling with opaque `workspaces::error::Error` type: https://github.com/near/workspaces-rs/pull/149
 - Require `#[must_use]` on the Execution value returned by `transact()`: https://github.com/near/workspaces-rs/pull/150
   - Added `ExecutionFinalResult`, `ExecutionResult`, `ExecutionSuccess` and `ExecutionFailure` types
@@ -11,8 +30,9 @@
   - Added `unwrap()` to not care about Err variant in `ExecutionResult`s
 
 ### Changed
-- Renamed CallExecution* types: https://github.com/near/workspaces-rs/pull/150
-  - Renamed `CallExecution`` to `Execution`
+
+- Renamed CallExecution\* types: https://github.com/near/workspaces-rs/pull/150
+  - Renamed ` CallExecution`` to  `Execution`
   - Renamed `CallExecutionDetails` to `ExecutionFinalResult`
 - `args_json` and `args_borsh` no longer return `Result`s and are deferred till later when `transact()`ed: https://github.com/near/workspaces-rs/pull/149
 - API changes from removing `worker` parameter from function calls: https://github.com/near/workspaces-rs/pull/181
@@ -20,54 +40,65 @@
   - `workspaces::prelude::*` import no longer necessary, where we no longer able to import `workspaces::prelude::DevAccountDeployer` directly.
 
 ### Removed
+
 - Removed impls from exection result: https://github.com/near/workspaces-rs/pull/150
   - Removed `impl<T> From<CallExecution<T>> for Result<T>`
   - Removed `impl From<FinalExecutionOutcomeView> for CallExecutionDetails`
 - No longer require `worker` to be passed in for each transaction: https://github.com/near/workspaces-rs/pull/181
 
 ### Fixed
+
 - Gas estimation issue resolved with latest sandbox node (Aug 29, 2022): https://github.com/near/workspaces-rs/pull/188
 - Fixed parallel tests, where calling into the same contract would require waiting on a previous call: https://github.com/near/workspaces-rs/pull/173
 
 ## [0.4.1] - 2022-08-16
 
 ### Added
+
 - Derive `Eq` on `AccountDetails` type: https://github.com/near/workspaces-rs/pull/177/files
 
 ### Fixed
+
 - Fix macOS non-deterministic overflow error when starting up sandbox: https://github.com/near/workspaces-rs/pull/179
 
 ## [0.4.0] - 2022-07-20
 
 ### Added
+
 - Mac M1 Support: https://github.com/near/workspaces-rs/pull/169
 - Added `Account::secret_key` to grab the account's secret key: https://github.com/near/workspaces-rs/pull/144
 - `Debug`/`Clone` impls for `Account`/`Contract`, and `Debug` for `Worker`: https://github.com/near/workspaces-rs/pull/167
 - `ExecutionOutcome::tokens_burnt` is now available: https://github.com/near/workspaces-rs/pull/168
 
 ### Fixed
+
 - internally no longer creating a new RPC client per call: https://github.com/near/workspaces-rs/pull/154
 - upped near dependencies to fix transitive vulnerabilities: https://github.com/near/workspaces-rs/pull/169
 
 ### Changed
+
 - Default sandbox version is now using commit hash master/13a66dda709a4148f6395636914dca2a55df1390 (July 18, 2022): https://github.com/near/workspaces-rs/pull/169
 
 ## [0.3.1] - 2022-06-20
 
 ### Added
+
 - Raw bytes API similar to `json`/`borsh` calls: https://github.com/near/workspaces-rs/pull/133/files
 - Expose `types` module and added `SecretKey` creation: https://github.com/near/workspaces-rs/pull/139
 
 ### Fixed
+
 - If sandbox gets started multiple times, short circuit it early on: https://github.com/near/workspaces-rs/pull/135
 - Fix short timeouts on connecting to RPC for macos with custom env variable to specify timeout if needed: https://github.com/near/workspaces-rs/pull/143
 
 ## [0.3.0] - 2022-05-10
 
 ### Added
+
 - Added betanet support https://github.com/near/workspaces-rs/pull/116
 
 ### Changed
+
 - Updated default sandbox version to `97c0410de519ecaca369aaee26f0ca5eb9e7de06` commit of nearcore to include 1.26 protocol changes https://github.com/near/workspaces-rs/pull/134
 
 - Exposed `CallExecutionDetails::raw_bytes` API: https://github.com/near/workspaces-rs/pull/133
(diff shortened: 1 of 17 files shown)


workspaces/src/rpc/query.rs
```diff
@@ -0,0 +1,361 @@
...
+
+impl ProcessQuery for GasPrice {
+    type Method = methods::gas_price::RpcGasPriceRequest;
+    type Output = Balance;
+
+    fn into_request(self, block_ref: BlockReference) -> Result<Self::Method> {
+        let block_id = match block_ref {
+            BlockReference::BlockId(block_id) => Some(block_id),
+            BlockReference::Finality(_finality) => None,
+            _ => panic!(""),
```
austinabell, line 352: Why is this necessary that can't be overcome with a local type? If it's necessary, can you at least add a message?

workspaces/src/worker/impls.rs
```diff
@@ -129,11 +157,17 @@ where
...
+            self.client(),
+            ViewAccount {
+                account_id: account_id.clone(),
+            },
+        )
+    }
+
+    pub fn gas_price(&self) -> Query<'_, GasPrice> {
+        Query::new(self.client(), GasPrice)
     }
```
austinabell, line 171: curious why all methods have the `view_` prefix except this one?

workspaces/tests/deploy.rs
```diff
@@ -43,7 +43,11 @@ async fn test_dev_deploy() -> anyhow::Result<()> {
         .await?
         .into_result()?;
 
-    let actual: NftMetadata = contract.view("nft_metadata", Vec::new()).await?.json()?;
+    let actual: NftMetadata = contract
+        .view("nft_metadata")
+        .args(Vec::new())
```
austinabell, line 48:
```suggestion
```

Isn't that the default? If this doesn't work, shouldn't it?

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
 
 ## [Unreleased]
 
+### Added
+
+- `view_*` asynchronous builders have been added which provides being able to query from a specific [`BlockReference`]()
```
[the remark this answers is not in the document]
ChaoticTempest: whoops

CHANGELOG.md
```diff
@@ -155,8 +174,12 @@
 
 - Fix race condition when installing sandbox and running multiples tests at the same time. https://github.com/near/workspaces-rs/pull/46
 
+<<<<<<< HEAD
+[unreleased]: https://github.com/near/workspaces-rs/compare/0.5.0...HEAD
+=======
 [unreleased]: https://github.com/near/workspaces-rs/compare/0.6.0...HEAD
 [0.6.0]: https://github.com/near/workspaces-rs/compare/0.5.0...0.6.0
+>>>>>>> 3d9a76a593470643342d7acb8f2903e941d901b1
```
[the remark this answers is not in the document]
ChaoticTempest: aww man, can't believe I missed this one

workspaces/src/rpc/query.rs
ChaoticTempest: oops, that panic should've returned an error. Not sure what you mean by local type here. Something like `GasPrice { block_id: ... }`? All this is doing is reusing the common `block_ref` in every `Query` for the block_id portion. The finality is default case when no `block_hash` or `block_height` function is used, but definitely should leave a comment saying why this is the case here 👍

workspaces/src/worker/impls.rs
ChaoticTempest: I just went with the naming here in https://docs.near.org/api/rpc/gas to stay consistent with the general naming, but since this is code, and people can lookup methods name at every keystroke with an IDE, it might be easier just to name it `view_gas_price` for that discoverability while coding. Do you have any opinions on this on which way it should lean?

workspaces/src/rpc/query.rs
```diff
@@ -0,0 +1,361 @@
...
+pub struct ViewAccessKey {
+    pub(crate) account_id: AccountId,
+    pub(crate) public_key: PublicKey,
+}
+
+pub struct ViewAccessKeyList {
+    pub(crate) account_id: AccountId,
+}
+
+pub struct GasPrice;
```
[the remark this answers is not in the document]
ChaoticTempest: Done, let me know if that module level doc doesn't do it

workspaces/src/rpc/query.rs
austinabell: I just mean, there is no way in the code to construct this, so we could just restrict by having an enum that doesn't contain that variant to avoid it completely by the type system rather than returning an error?

workspaces/src/worker/impls.rs
austinabell: I didn't realize these method names were matching the RPC names. I think it's fine to maintain parity with those names 👍

examples/src/status_message.rs
```diff
@@ -18,14 +18,10 @@ async fn main() -> anyhow::Result<()> {
...
-        .view(
-            "get_status",
-            json!({
-                "account_id": contract.id(),
-            })
-            .to_string()
-            .into_bytes(),
-        )
+        .view("get_status")
+        .args_json(json!({
```
matklad, line 22: drive by, but some examples use `serde_json::json`, and some just `json`. I suggest unifying them all to `json`. Consistency is imporatnt for examples, because they get copy-pasted into downstream projects and set the initial patterns.

workspaces/src/operations.rs
```diff
@@ -24,17 +24,21 @@ const MAX_GAS: Gas = 300_000_000_000_000;
...
+    pub(crate) name: T,
+    pub(crate) args: Result<Vec<u8>>,
+    pub(crate) deposit: Balance,
+    pub(crate) gas: Gas,
 }
 
-impl<'a> Function<'a> {
+pub type Function<'a> = FunctionArgs<&'a str>;
+/// Owned version of [`Function`], whereby a lifetime is not attached.
+pub type FunctionOwned = FunctionArgs<String>;
```
matklad, line 36: It doesn't look like the non-owned type pulls its weight. `Function` will eventually lead to a network call, a string allocation is nothing compared to that.

matklad (commented):
IntoFuture stuff looks correct to me. 

What caught my eye is that we are enduring some pain (and even causing the users paint) to be generic over `&str` vs `String`. Did we measured that using `&str` gives us a measurable performance improvements? If we did not, I suggest sticking with owned `String` and clone, and removing extra generics from the API.

workspaces/src/rpc/query.rs
ChaoticTempest: Ahh, I see what you mean. I did something like that initially, with another associated type `Input` in the `ProcessQuery` trait, but it just complicated things, when `GasPrice` was the only extra case. Like, we'd have to add in a lot more boilerplate code to be specific about just having either a `BlockReference` vs just a `BlockId` for gas price, so I opted not to do that for now until we have at least 2 or 3 that depended on that behavior. It's all internal anyways, so wouldn't be exposed if we decide to change it later

workspaces/Cargo.toml
```diff
@@ -18,6 +18,7 @@ bs58 = "0.4"
 cargo_metadata = { version = "0.14.2", optional = true }
 chrono = "0.4.19"
 dirs = "3.0.2"
+futures = "0.3"
```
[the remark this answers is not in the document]
ChaoticTempest: originally used futures a bit more than just importing `BoxFuture`, but seems like that's the only import, I directly pulled it into `crate::rpc::BoxFuture` so that we don't need this import

workspaces/src/operations.rs
ChaoticTempest: That's true. `Function<'a>` is leaning towards micro optimization, but also makes the user experience a bit rougher if they wanted to store these `Function` objects in their own struct. I'll just move to the owned version unless there was something else. @austinabell if you had any opinions about this

workspaces/src/rpc/query.rs
```diff
@@ -0,0 +1,393 @@
...
+            // User provided input via `block_hash` or `block_height` functions.
+            BlockReference::BlockId(block_id) => Some(block_id),
+            // default case, set by `Query` struct via BlockReference::latest.
+            BlockReference::Finality(_finality) => None,
+            // Should not be reachable, unless code got changed.
+            other => {
+                return Err(RpcErrorCode::QueryFailure.message(format!(
+                    "Invalid block reference provided to gas price: {other:?}"
+                )))
+            }
```
austinabell, line 384: nit: would be nice if the other case is handled specifically, so that if another variant is added in the future, it won't default to erroring

workspaces/src/rpc/query.rs
```diff
@@ -0,0 +1,393 @@
+//! This module defines a bunch of internal types used solely for querying into RPC
+//! methods to retrieve info about what's on the chain. Note that the types defined
+//! exposed as-is for users to reference in their own functions or structs as needed.
+//! These types cannot be created outside of workspaces. To use them, refer to
+//! surface level types like [`Account`], [`Contract`] and [`Worker`].
+//!
+//! For example, to query into downloading
+//! contract state:
+//! ```ignore
```
austinabell, line 9: would be nice if this wasn't ignored so that the interface is run through doc tests and this doesn't get outdated if changes, nbd though

workspaces/src/operations.rs
```diff
@@ -236,19 +236,19 @@ impl<'a> Transaction<'a> {
...
+pub struct CallTransaction<'a> {
     worker: &'a Worker<dyn Network>,
     signer: InMemorySigner,
     contract_id: AccountId,
-    function: Function<'b>,
+    function: Function,
 }
 
-impl<'a, 'b> CallTransaction<'a, 'b> {
+impl<'a, 'b> CallTransaction<'a> {
```
austinabell, line 246: isn't `'b` unused?

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+    worker.view_state("account_id", Some(prefix)).await?;
+    // is now
+    worker.view_state("account_id")
+        .prefix(prefix)
+        .await?;
+    // if prefix was `None`, then simply delete the None argument.
+    ```
+  - `view` function changed to be a builder, and no longer take in `args` as a parameter. It instead has been moved to the builder side.
+  - Changed `Worker::view_latest_block` to `Worker::view_block` as the default behavior is equivalent.
+  - `operations::Function` type no longer takes a lifetime parameter.
```
austinabell, line 23: Maybe worth also noting `CallTransaction` has one less lifetime for the same reason

austinabell (approved):

workspaces/src/operations.rs
ChaoticTempest: I'm surprised there weren't any warnings for this both from rustc and clippy

PR #220, opened 2022-10-14
feat: workspaces::sandbox() can connect to user provided sandbox node 

ChaoticTempest:
This adds a `Builder` to the return type of `workspaces::sandbox()` which is another async builder so that users can parameterize what RPC url they want to point to if they have a different RPC node pointing to that network. For this PR, only sandbox was done, and will add `testnet` and `mainnet` after since those are pretty trivial in comparison.

Sandbox now basically has the ability to connect to a user's own locally spawned sandbox node.

### DIFF
--- a/examples/manually-spawned-sandbox.md
+++ b/examples/manually-spawned-sandbox.md
@@ -1,0 +1,28 @@
+# Utilizing custom sandbox node
+
+This example will show us how to spin up a sandbox node of our own choosing. Follow the guide in https://github.com/near/sandbox to download it. This is mainly needed if a user wants to manage their own node and/or not require each test to spin up a new node each time.
+
+Then initialize the chain via `init` and run it:
+
+```sh
+near-sandbox --home ${MY_HOME_DIRECTORY} init
+near-sandbox --home ${MY_HOME_DIRECTORY} run
+```
+
+This will launch the chain onto `localhost:3030` by default. The `${MY_HOME_DIRECTORY}` is a path of our choosing here and this will be needed when running the workspaces code later on. In the following example, we had it set to `/home/user/.near-sandbox-home`.
+
+In workspaces, to connect to our manually launched node, all we have to do is add a few additional parameters to `workspaces::sandbox()`:
+
+```rs
+#[tokio::main]
+fn main() {
+    let worker = workspaces::sandbox()
+        .rpc_addr("http://localhost:3030")
+        .home_dir("/home/user/.near-sandbox-home")
+        .await?;
+
+    Ok(())
+}
+```
+
+Then afterwards, we can continue performing our tests as we normally would if workspaces has spawned its own sandbox process.
--- a/workspaces/src/network/builder.rs
+++ b/workspaces/src/network/builder.rs
@@ -61,17 +61,9 @@ impl<'a, T> NetworkBuilder<'a, T> {
     /// node than the default one provided by near.org. This enables certain features that
     /// the default node doesn't provide such as getting beyond the data cap when downloading
     /// state from the network.
-    /// Example:
-    /// ```
-    /// async fn connect_custom() -> anyhow::Result<()> {
-    ///     let worker = workspaces::testnet()
-    ///         .rpc_addr("https://rpc.testnet.near.org")
-    ///         .await?;
-    ///     Ok(())
-    /// }
-    /// ```
     ///
-    /// Note that, for sandbox, we also are required to specify `home_dir`
+    /// Note that, for sandbox, we are required to specify `home_dir` as well to connect to
+    /// a manually spawned sandbox node.
     pub fn rpc_addr(mut self, addr: &str) -> Self {
         self.rpc_addr = Some(addr.into());
         self
--- a/workspaces/src/network/builder.rs
+++ b/workspaces/src/network/builder.rs
@@ -1,0 +1,92 @@
+use std::future::{Future, IntoFuture};
+use std::marker::PhantomData;
+use std::path::PathBuf;
+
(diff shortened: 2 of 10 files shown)


workspaces/src/network/builder.rs
```diff
@@ -0,0 +1,81 @@
+use std::future::{Future, IntoFuture};
+use std::marker::PhantomData;
+use std::path::PathBuf;
+
+use crate::network::Sandbox;
+use crate::{Network, Worker};
+
+pub type BoxFuture<'a, T> = std::pin::Pin<Box<dyn Future<Output = T> + Send + 'a>>;
```
itegulov, line 8: Interesting. Feels a little strange to do an additional allocation here, but I am guessing you can't just use `impl Future<...>` because the type is not concrete?

itegulov, line 8: Doesn't really matter, but shouldn't this be `pub(crate)`?

examples/manually-spawned-sandbox.md
```diff
@@ -0,0 +1,28 @@
+# Utilizing custom sandbox node
+
+This example will show us how to spin up a sandbox node of our own choosing. Follow the guide in https://github.com/near/sandbox to download it.
```
itegulov, line 3: I would also provide reasoning why a dev would want to spin their own node

workspaces/src/network/server.rs
```diff
@@ -1,49 +1,95 @@
...
             return Err(SandboxErrorCode::AlreadyStarted.into());
         }
 
-        info!(target: "workspaces", "Starting up sandbox at localhost:{}", self.rpc_port);
-        let home_dir = Sandbox::home_dir(self.rpc_port);
-
+        let rpc_port = self.rpc_addr.port().ok_or_else(|| {
+            SandboxErrorCode::InitFailure.message("No port for sandbox: bug in code")
+        })?;
+        info!(target: "workspaces", "Starting up sandbox at localhost:{}", self.rpc_addr.port().unwrap());
```
itegulov, line 79:
```suggestion
        info!(target: "workspaces", "Starting up sandbox at localhost:{}", rpc_port);
```

itegulov (commented):
Would also be nice to add a test for this

workspaces/src/network/builder.rs
ChaoticTempest: it's because `impl Future` in a type alias is unstable 😞

ChaoticTempest:
Rebased this with latest changes on workspaces. @itegulov let me know how it looks when you get the chance

workspaces/src/network/server.rs
```diff
@@ -111,22 +148,20 @@ impl Drop for SandboxServer {
...
 
+        let rpc_port = self.rpc_port();
         let child = self.process.as_mut().unwrap();
 
         info!(
             target: "workspaces",
-            "Cleaning up sandbox: port={}, pid={}",
-            self.rpc_port,
+            "Cleaning up sandbox: port={:?}, pid={}",
+            rpc_port,
```
miraclx, line 157: For messages like these, isn't just a port lossy? Especially since custom sandbox Urls could technically have no port. So this would just be `None`.

workspaces/src/network/server.rs
```diff
@@ -111,22 +148,20 @@ impl Drop for SandboxServer {
...
             child.id()
         );
 
         child
             .kill()
             .map_err(|e| format!("Could not cleanup sandbox due to: {:?}", e))
             .unwrap();
-
-        // Unlock the ports just in case they have not been preemptively done.
-        self.unlock_lockfiles().unwrap();
```
miraclx, line 129: Was this intentional? This should be safe, no?

workspaces/src/network/server.rs
```diff
@@ -111,22 +148,20 @@ impl Drop for SandboxServer {
             return;
         }
 
+        let rpc_port = self.rpc_port();
         let child = self.process.as_mut().unwrap();
```
miraclx, line 152: I believe this would panic for custom sandbox urls.

miraclx (changes requested):

workspaces/src/network/server.rs
ChaoticTempest: For custom sandbox URLs, there's no process spawned, so `child` would be None here and early return in the above:
```rust
        if self.process.is_none() {
            return;
        }
```

workspaces/src/network/server.rs
ChaoticTempest: good point, this was old messaging that I forgot and didn't notice to change. Thanks for pointing this out

workspaces/src/network/server.rs
ChaoticTempest: this was indeed intentional, since I noticed that fs2 actually already drops them on `Drop`s so no need to add it here

workspaces/src/network/server.rs
miraclx: Ah, that code path was collapsed in the diff. Looks good.

miraclx: Wanna try this at the top instead?

```rust
let child = match self.process.take() {
    Some(child) => child,
    None => return,
};
```

nabeel99:
hey @ChaoticTempest , as I stated on discord , I am trying to integrate workspaces local/sandbox tests with the kurtosis local environment, would it be possible to use this branch to connect to it ? if not what roadblock would prevent me from doing it ?

itegulov:
@miraclx any objections to merging this?

miraclx (approved):
None at all. LGTM

PR #221, opened 2022-10-18
feat: add `NetworkBuilder` impl for mainnet, betanet and testnet

ChaoticTempest:
Second part of https://github.com/near/workspaces-rs/pull/220

This implements the `NetworkBuilder` for testnet and mainnet networks so that these networks can customize the RPC urls to point to Pagoda ones or their own RPC. Also made, `RPC_URL` and `ARCHIVAL_URL` constants public, since users can make use of them if they want to.


ChaoticTempest:
Also, just added `Betanet` impl as well, since I forgot about it 😅

frol (approved):

ChaoticTempest:
The latest rust version stable tests are failing due to a contract getting compiled with 1.70, so merging this one in. Already tested the rest locally and everything else works

Issue #216, opened 2022-10-07
Add `transact_async` for asynchronous transactions


ChaoticTempest:
Currently, the flow for sending a transaction is by calling into `transact()`, which is a synchronous operation and blocks the next transaction from happening. This isn't ideal for users who want to have transactions that don't depend on the previous transaction.  So to alleviate this, additional APIs need to be expose so that users can broadcast an asynchronous transaction, while being able to track the status of it.

austinabell:
This could be closed now. I'm just commenting to question the naming. Something more intuitive to me would be a name like `broadcast` rather than `transact_async`. This naming seems like it's referring to the tx type at a Rust level. Wdyt?

ChaoticTempest:
So, underneath in the raw RPC call, it's actually called `broadcast_tx_commit` and `broadcast_tx_async` for `transact` and `transact_async` respectively. People might confuse `broadcast` with both those rpc calls and which one we're referring to. So, I don't think that's the better name here. I feel like `transact_async` conveys that we're trying to perform a transaction while having it be asynchronous, but I can see how it would get confused to be a certain Rust type instead. Not sure how to remedy that part but feels like that isn't too much of a concern from what I see. Or what is the expectation when you go and use `transact_async`?

> This could be closed now.

😅 I would've thought github closed this. It seems to sometimes do it and sometimes don't. Thanks for pointing that this is still open

PR #222, opened 2022-10-19
feat: Expose `transact async` for asynchronous transactions

ChaoticTempest:
This adds `transact_async` to `CallTransaction` and `Transaction` objects so users can call into committing transactions without having to wait for them to complete if their transactions are independent of each other. `TransactionStatus` type also added to check the status of the transaction afterwards and finally grab its result.

addresses https://github.com/near/workspaces-rs/issues/216

### DIFF
--- a/workspaces/src/operations.rs
+++ b/workspaces/src/operations.rs
@@ -226,7 +226,7 @@ impl<'a> Transaction<'a> {
         send_batch_tx_and_retry(self.client, &self.signer, &self.receiver_id, self.actions?).await
     }
 
-    /// Process the trannsaction, and return the result of the execution.
+    /// Process the transaction, and return the result of the execution.
     pub async fn transact(self) -> Result<ExecutionFinalResult> {
         self.transact_raw()
             .await
@@ -234,6 +234,13 @@ impl<'a> Transaction<'a> {
             .map_err(crate::error::Error::from)
     }
 
+    /// Send the transaction to the network to be processed. This will be done asynchronously
+    /// without waiting for the transaction to complete. This returns us a [`TransactionStatus`]
+    /// for which we can call into [`status`] and/or [`wait`] to retrieve info about whether
+    /// the transaction has been completed or not.
+    ///
+    /// [`status`]: TransactionStatus::status
+    /// [`wait`]: TransactionStatus::wait
     pub async fn transact_async(self) -> Result<TransactionStatus<'a>> {
         send_batch_tx_async_and_retry(self.client, &self.signer, &self.receiver_id, self.actions?)
             .await
@@ -324,6 +331,13 @@ impl<'a> CallTransaction<'a> {
             .map_err(crate::error::Error::from)
     }
 
+    /// Send the transaction to the network to be processed. This will be done asynchronously
+    /// without waiting for the transaction to complete. This returns us a [`TransactionStatus`]
+    /// for which we can call into [`status`] and/or [`wait`] to retrieve info about whether
+    /// the transaction has been completed or not.
+    ///
+    /// [`status`]: TransactionStatus::status
+    /// [`wait`]: TransactionStatus::wait
     pub async fn transact_async(self) -> Result<TransactionStatus<'a>> {
         send_batch_tx_async_and_retry(
             self.worker.client(),
@@ -416,10 +430,10 @@ impl<'a, 'b> CreateAccountTransaction<'a, 'b> {
     }
 }
 
-/// TransactionStatus object of an [`asynchronous transaction`]. Used to query into
-/// the transaction status of
+/// `TransactionStatus` object relating to an [`asynchronous transaction`] on the network.
+/// Used to query into the status of the Transaction for whether it has completed or not.
 ///
-/// [`asynchrouns transaction`]: https://docs.near.org/api/rpc/transactions#send-transaction-async
+/// [`asynchronous transaction`]: https://docs.near.org/api/rpc/transactions#send-transaction-async
 #[must_use]
 pub struct TransactionStatus<'a> {
     client: &'a Client,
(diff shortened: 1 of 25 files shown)


workspaces/tests/parallel_transactions.rs
```diff
@@ -38,3 +40,42 @@ async fn test_parallel() -> anyhow::Result<()> {
...
+        statuses.push_back(status);
+    }
+
+    // Retry checking the statuses of all transactions until the queue is empty
+    // with all transactions completed.
+    while let Some(status) = statuses.pop_front() {
+        if let Err(_err) = status.status().await {
+            statuses.push_back(status);
+        }
+    }
```
austinabell, line 69: Feels like it would be much more ergonomic if the type returned from `transact_async` was a future so that you can simply use something like `try_join_all`. Also somewhat confusing what this test is doing because there is no need for the polling to cycle if you are waiting on all futures to complete.

What was the reasoning for not using a future type for the `transact_async` return?

workspaces/tests/parallel_transactions.rs
```diff
@@ -38,3 +40,42 @@ async fn test_parallel() -> anyhow::Result<()> {
...
+
+    // Check the final set message. This should be "j" due to the ordering of the queue.
+    let final_set_msg = account
+        .call(contract.id(), "get_status")
+        .args_json(json!({ "account_id": account.id() }))
+        .view()
+        .await?
+        .json::<String>()?;
```
austinabell, line 77: This test doesn't check that all the transactions executed, just that the last one did. Maybe worth adding in a check to check the nonce increase or the count of successful receipts?

workspaces/src/operations.rs
```diff
@@ -398,3 +434,61 @@ impl<'a, 'b> CreateAccountTransaction<'a, 'b> {
...
+        Self {
+            client,
+            sender_id: id,
+            hash: CryptoHash(hash.0),
+        }
+    }
+
+    /// Checks the status of the transaction. If an `Err` is returned, then the
+    /// transaction has not completed yet.
+    pub async fn status(&self) -> Result<ExecutionFinalResult> {
```
austinabell, line 464: Isn't this misleading that a non-finalized result would be treated as a result? Wouldn't you want something like `Result<Option<..>>` or some other way to indicate the `Poll::NotReady` state essentially?

workspaces/tests/parallel_transactions.rs
```diff
@@ -38,3 +40,42 @@ async fn test_parallel() -> anyhow::Result<()> {
...
+            .await?;
+        statuses.push_back(status);
+    }
+
+    // Retry checking the statuses of all transactions until the queue is empty
+    // with all transactions completed.
+    while let Some(status) = statuses.pop_front() {
+        if let Err(_err) = status.status().await {
+            statuses.push_back(status);
+        }
```
austinabell, line 68: Noting that if there is an unrelated error, this will just infinite loop, right? Maybe best to avoid as debugging a hanging test might be painful in the future

workspaces/tests/parallel_transactions.rs
ChaoticTempest: yeah it could just be a `Future`, but I feel like other information like the crypto hash and account-id, someone might want to reference for themselves. But in that case, let's have this `TransactionStatus` impl `Future` then so we can get the same result

workspaces/src/operations.rs
ChaoticTempest: true, we should probably have a different variant than just `Result`. I was just parroting what we had in the client side, but definitely better for users if we had a custom enum that portrayed this poll-like state

workspaces/tests/parallel_transactions.rs
ChaoticTempest: good idea and added

workspaces/tests/parallel_transactions.rs
ChaoticTempest: With the `TransactionPoll::{Pending, Error}` variants, this shouldn't be an issue

workspaces/src/operations.rs
```diff
@@ -398,3 +434,61 @@ impl<'a, 'b> CreateAccountTransaction<'a, 'b> {
...
+    }
+
+    /// Wait till completion of the transaction.
+    pub async fn wait(self) -> ExecutionFinalResult {
+        loop {
+            if let Ok(val) = self.status().await {
+                break val;
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(300)).await;
```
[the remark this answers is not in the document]
ChaoticTempest: users can do something `tokio::time::timeout()` to control the extend future if they need to, but seems like every 300ms seems reasonable for now

workspaces/src/operations.rs
```diff
@@ -393,3 +433,90 @@ impl<'a, 'b> CreateAccountTransaction<'a, 'b> {
...
+        Self {
+            client,
+            sender_id: id,
+            hash: CryptoHash(hash.0),
+        }
+    }
+
+    /// Checks the status of the transaction. If an [`TransactionPoll::Pending`] is returned, then the
+    /// transaction has not completed yet.
+    pub async fn status(&self) -> TransactionPoll<ExecutionFinalResult> {
```
austinabell, line 463: I wonder if using `Result<std::task::Poll<T>>` instead of a custom type might be preferred for ergonomics?

This seems like it might not be that ergonomic to use because they would have to match on the error if they wanted to just fail if it's hit.

What's the reason for introducing the new type `TransactionPoll`?

ChaoticTempest: `std::task::Poll` didn't have an error vairant, so went with a custom type, but then again, `Poll` has a couple impls that are nice to have and we don't repeat ourselves. Switched `Result<std::task::Poll<T>>` instead. Thanks for the suggestion

austinabell (approved):

Issue #194, opened 2022-09-20
Ability to create Workspace errors


joshuajbouw:
We aim to build a wrapper around the `Contract` struct for the Aurora EVM. What would be a nice feature is if we can create our own workspace `Error`s, specifically the `ErrorRepr::Simple`, `ErrorRepr::Message`, and `ErrorRepr::Custom`. The benefit of this is that those that use our EVM wrapper will be able to have the ability also to use the same `Error`s as `Workspace`. 

I don't see why anyone would ever need access to the more detailed errors which include `ExecutionFailure`, as those should only come from the execution itself.

### More than one error type (more work, best solution)

The errors could be split into more specific error returns depending on the context. For example, one each for data conversion errors, rpc, execution, sandbox, and IO.

### Remove `pub(crate)` in some cases (less work)

The `pub(crate)` could be removed such as in `Error::custom`, `Error::message`, and `Error::simple`.

ChaoticTempest:
> The errors could be split into more specific error returns depending on the context. For example, one each for data conversion errors, rpc, execution, sandbox, and IO.

Originally was split into their own specific errors, but had a lot of maintenance burden with not much real value add, since at the end of the day, errors were getting merged into the higher level `workspaces::Error` object anyways. So it's much simpler to have this one singular `Error` type where we can internally just scale the `ErrorKind` vairants without much further work

ChaoticTempest:
resolved with latest 0.7.0 version

PR #224, opened 2022-10-25
chore: Expose error creation methods

ChaoticTempest:
addresses https://github.com/near/workspaces-rs/issues/194.

Thought it would be nice to have these be exposed since the errors are in good shape to be used at higher levels if users want to make use of them to build their own errors. Also added the `Other` variant to `ErrorKind` since users can have different categories of errors other than the ones defined for their higher level workspaces libraries. Any objections to adding this?

### DIFF
--- a/workspaces/src/error/impls.rs
+++ b/workspaces/src/error/impls.rs
@@ -38,7 +38,7 @@ impl Error {
     /// Construct a workspaces [`Error`] with the full details of an error which includes
     /// the internal error it references, the custom error message with further context
     /// and the [`ErrorKind`] that represents the category of error.
-    pub fn full<T, E>(kind: ErrorKind, msg: T, error: E) -> Self
+    pub(crate) fn full<T, E>(kind: ErrorKind, msg: T, error: E) -> Self
     where
         T: Into<Cow<'static, str>>,
         E: Into<Box<dyn std::error::Error + Send + Sync>>,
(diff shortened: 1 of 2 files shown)


workspaces/src/error/impls.rs
```diff
@@ -35,7 +35,10 @@ impl Error {
         }
     }
 
-    pub(crate) fn full<T, E>(kind: ErrorKind, msg: T, error: E) -> Self
+    /// Construct a workspaces [`Error`] with the full details of an error which includes
+    /// the internal error it references, the custom error message with further context
+    /// and the [`ErrorKind`] that represents the category of error.
+    pub fn full<T, E>(kind: ErrorKind, msg: T, error: E) -> Self
```
austinabell, line 41: Just going to share some low-confidence opinions that having the semantics of error and message separate seems unnecessarily complex and having this interface now be public seems like it might be a pain to maintain in the future.

If you're confident this is useful and these should all be exposed, go for making this change, I just noticed that there is no actual usage of the downcasting through `into_inner` of `Error` and exposing all of these variants will make it even harder to switch later.

austinabell (approved):

workspaces/src/error/impls.rs
ChaoticTempest: I see, how about just expose only custom/message errors then. There's only a couple cases where full is useful in workspaces itself where we just provide some extra context about the error. So let's not expose it for now until there's a community need

PR #231, opened 2022-11-03
fix: Use public keys for cached nonces

ChaoticTempest:
Moves the HashMap key from `AccountId` to `PublicKey`, since we want to associate it based on per-access-key instead of per account

Also, added debug/clone impls for the types that were added in recent PRs

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -333,15 +333,15 @@ impl Client {
 
 pub(crate) async fn access_key(
     client: &Client,
-    account_id: &near_primitives::account::id::AccountId,
-    public_key: &near_crypto::PublicKey,
+    account_id: near_primitives::account::id::AccountId,
+    public_key: near_crypto::PublicKey,
 ) -> Result<(AccessKeyView, CryptoHash)> {
     let query_resp = client
         .query(&methods::query::RpcQueryRequest {
             block_reference: Finality::None.into(),
             request: QueryRequest::ViewAccessKey {
-                account_id: account_id.clone(),
-                public_key: public_key.clone(),
+                account_id,
+                public_key,
             },
         })
         .await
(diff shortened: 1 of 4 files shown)


austinabell:
Actually, what you want is a combination of the two. You can have the same public key as an access key on different accounts

workspaces/src/rpc/client.rs
```diff
@@ -333,15 +333,15 @@ impl Client {
...
+    public_key: &near_crypto::PublicKey,
 ) -> Result<(AccessKeyView, CryptoHash)> {
     let query_resp = client
         .query(&methods::query::RpcQueryRequest {
             block_reference: Finality::None.into(),
             request: QueryRequest::ViewAccessKey {
-                account_id,
-                public_key,
+                account_id: account_id.clone(),
+                public_key: public_key.clone(),
```
austinabell, line 344: I don't see why this is changed. If this requires ownership, why not leave as an owned param and let the caller clone or not if needed?

workspaces/src/rpc/client.rs
```diff
@@ -373,22 +373,22 @@ async fn cached_nonce(nonce: &AtomicU64, client: &Client) -> Result<(CryptoHash,
 /// into contention with others.
 async fn fetch_tx_nonce(
     client: &Client,
-    account_id: AccountId,
-    public_key: near_crypto::PublicKey,
+    cache_key: &(AccountId, near_crypto::PublicKey),
```
austinabell, line 376: same with this, feels like you avoid a clone if you just have this be an owned value and not take a ref outside?

austinabell (approved):
my comment is opinionated and doesn't matter, feel free to ignore

workspaces/src/rpc/client.rs
ChaoticTempest: Same with the other one with trying to optimize a little bit. But my thought was that since `fetch_tx_nonce` only really needs to clone on writing the nonce with key initially, it'd be better to not give it owned values everytime. Especially more so when `send_batch_tx_and_retry` can take the actual owned value later.

The most recent commit I had should make better due of this situation though. Moved the initial creation of the cache_key outside of the retry loop, so we can just always take a reference to it, and removed the need to take ownership of the cache_key when merely just deleting the entry from the nonce map.

ChaoticTempest:
merging this since this also fixes the clippy issues with rust 1.65 update

PR #232, opened 2022-11-07
WIP: don't lose error context

matklad:
This is more of a bug report than a PR, I don't plan to make this production ready :)

It seems we are loosing quite a bit of error context somewhere.

Before this PR:

```
λ cargo run --example nft
Error: No such file or directory (os error 2)
```

After this PR: 

```
λ cargo run --example nft
Error: Could not initialize sandbox node

Caused by:
    No such file or directory (os error 2)
```

Note that even improved output is quite suboptimal, as it doesn't tell _which_ file or directory is not found, but the roots of that are deep in the `sandbox-util` crate

### DIFF
--- a/workspaces/src/error/mod.rs
+++ b/workspaces/src/error/mod.rs
@@ -19,7 +19,7 @@ pub enum ErrorKind {
     #[error("Execution")]
     Execution,
     /// An error having to do with running sandbox.
-    #[error("Sandbox({0})")]
+    #[error("{0}")]
     Sandbox(#[from] SandboxErrorCode),
     /// An error from performing IO.
     #[error("IO")]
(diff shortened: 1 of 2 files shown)


workspaces/src/error/mod.rs
```diff
@@ -19,7 +19,7 @@ pub enum ErrorKind {
     #[error("Execution")]
     Execution,
     /// An error having to do with running sandbox.
-    #[error("Sandbox({0})")]
+    #[error("{0}")]
```
matklad, line 22: Note: `error` is Display, it should not use the `TypeName(contents)` form, that one is only for Debug.

matklad:
https://github.com/near/sandbox/pull/55 -- the roots

ChaoticTempest:
thanks for pointing this out! I'll merge this one in when the sandbox side gets updated to reflect the `with_context` changes

PR #234, opened 2022-11-09
chore: expose `view_chunk`

ChaoticTempest:
This adds `worker.view_chunk` method for users to call into. Internally, uses `QueryChunk`, since `Query` is more for block reference related queries and nothing else uses `ChunkReference` when making queries.

Also adds a new `various_queries` example to showcase querying both chunk and block data, while being able to query for old contract state data as well.

### DIFF
--- a/workspaces/src/types/chunk.rs
+++ b/workspaces/src/types/chunk.rs
@@ -64,11 +64,6 @@ impl From<ChunkView> for Chunk {
 }
 
 impl Chunk {
-    /// The author's [`AccountId`] relating to the creation of this chunk.
-    pub fn author(&self) -> &AccountId {
-        &self.author
-    }
-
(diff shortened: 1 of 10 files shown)


workspaces/src/types/chunk.rs
```diff
@@ -0,0 +1,81 @@
...
+            },
+        }
+    }
+}
+
+impl Chunk {
+    /// The author's [`AccountId`] relating to the creation of this chunk.
+    pub fn author(&self) -> &AccountId {
+        &self.author
+    }
```
itegulov, line 70: Any reason to expose this if it's already a public field?

ChaoticTempest: ahh, yeah doesn't need to be exposed like this. Originally was a private field that I exposed later

PR #236, opened 2022-11-10
fix: improve err message

ChaoticTempest:
based on https://github.com/near/workspaces-rs/pull/232

Adds in the sandbox error improvement

### DIFF
--- a/workspaces/src/error/mod.rs
+++ b/workspaces/src/error/mod.rs
@@ -13,7 +13,7 @@ use crate::result::ExecutionFailure;
 #[non_exhaustive]
 pub enum ErrorKind {
     /// An error occurred while performing an RPC request.
-    #[error("Rpc({0})")]
+    #[error("{0}")]
     Rpc(#[from] RpcErrorCode),
     /// An error occurred while processing a transaction.
     #[error("Execution")]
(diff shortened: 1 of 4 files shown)


workspaces/src/error/mod.rs
```diff
@@ -38,15 +38,17 @@ enum ErrorRepr {
...
     },
-    #[error("{error}")]
+    #[error("{kind}")]
     Custom {
         kind: ErrorKind,
+        #[source]
         error: Box<dyn std::error::Error + Send + Sync>,
     },
-    #[error("{message}: {error}")]
+    #[error("{message}")]
```
austinabell, line 47: Curious in these why for `Custom` the error is not included in this message (does source automatically include it?), and for `Full`, why the error kind is ignored

austinabell (approved):
TIL about the source attribute

workspaces/src/error/mod.rs
ChaoticTempest: yup, `#[source]` creates a chain of error message when you go to display it. For `full`, didn't see a need to include it since the error message should for the most part tell you what kind of class of errors it it. Normally I don't think `kind` is that useful in error messages unless there's no other context to the error

ChaoticTempest:
hmm, looks like 4.1.1 breaks `compile_project`. Using `=4.0.0` fixes things but error messages aren't saying much

austinabell:
> Using `=4.0.0` fixes things but error messages aren't saying much

Prob worth opening an issue for that. The error message was in the output, but that was very cryptic to read

Issue #230, opened 2022-11-02
Replace ad-hoc build logic with `cargo-near`


itegulov:
I think the existing [cargo build logic](https://github.com/near/workspaces-rs/blob/main/workspaces/src/cargo/mod.rs) can be replaced with `cargo-near`'s public API (apart from being a CLI tool it also comes as a library). Less things to maintain => more better?

ChaoticTempest:
wouldn't this require users to install `cargo-near` themselves first? I guess we can add it to build.rs script, but can it be used directly as a rust library?

itegulov:
Yeah, it has a `[lib.rs](https://github.com/near/cargo-near/blob/main/cargo-near/src/lib.rs)`, so it should be able to use it directly as a regular library. The API might be awkward though, we haven't paid much attention to its "cli-as-a-library" design.

ChaoticTempest:
This should also alleviate the error message problem when unable to compile with `cargo build` shown in https://github.com/near/workspaces-rs/pull/236

ghimire007:
@frol As per our offline discussion, I would like work in this issue.

PR #238, opened 2022-11-10
fix: remove unused import of metadata

austinabell:
metadata was imported but not used previously, was removed in 4.1.


ChaoticTempest (approved):
ahh. Hopefully when workspaces switches to https://github.com/near/workspaces-rs/issues/230, the error messages could be more clear what went wrong

Issue #237, opened 2022-11-10
Sandbox shouldn't require setting a very high ulimit to run


austinabell:
Currently, there are the errors that require setting ulimit to run the node for tests:

```
Caused by:
    Hard limit for the number of open files (NOFILE) is too low (10240).  At least 11000 is required.  Set ‘ulimit -Hn’ accordingly and restart the node.', neard/src/cli.rs:423:22
```

This is an unreasonable expectation for users of workspaces and will be a frictional point. We should not have this same requirement for running a sandbox node for testing.

ChaoticTempest:
https://github.com/near/workspaces-rs/pull/241 has the potential fix based purely on how nearcore asserts that error. Haven't tested it below the bounds since I'm not sure how to configure it on macos, but let me know if you're able to test it

PR #241, opened 2022-11-15
fix: ulimit error

ChaoticTempest:
Unsure if this actually fixes https://github.com/near/workspaces-rs/issues/237, but it should according to nearcore code. Still need to test this with ulimits set below certain bounds, but this also provides a way to manually override the max open files with an environment var.

We shouldn't need to configure an env var for every configurable in `config.json` though, but requires https://github.com/near/workspaces-rs/pull/220 to make it in first before we can offer easier ways to configure things (by passing in configs into the sandbox builder)


ChaoticTempest:
tested on Windows WSL2 and fixes the issue

workspaces/src/network/config.rs
```diff
@@ -37,20 +37,18 @@ fn overwrite(home_dir: &Path, value: Value) -> Result<()> {
     Ok(())
 }
 
-/// Limit how much nearcore/sandbox can receive per payload. The default set by nearcore is not
-/// enough for certain sandbox operations like patching contract state in the case of contracts
-/// larger than 10mb.
-fn max_sandbox_json_payload_size() -> Result<u64> {
-    let max_files = match std::env::var("NEAR_SANDBOX_MAX_PAYLOAD_SIZE") {
+/// Parse an environment variable or return a default value.
+fn parse_env_or<T>(env_var: &str, default: T) -> Result<T>
```
DavidM-D, line 41: A real nit, so feel free to ignore, but shouldn't this be:
```suggestion
fn parse_env<T>(env_var: &str, default: T) -> Option<Result<T>>
```
Then called as:
```rust
parse_env_or("NEAR_SANDBOX_MAX_FILES").unwrap_or(3000)
```
It just feels a bit more general and standard rust

DavidM-D (approved):
LGTM

workspaces/src/network/config.rs
miraclx: Or, alternatively, consider returning a `Result<Option<T>>`

so instead of

```rust
parse_env("NEAR_SANDBOX_MAX_FILES").unwrap_or(Ok(3000))?;
```

it'll be

```rust
parse_env("NEAR_SANDBOX_MAX_FILES")?.unwrap_or(3000);
```

subtle change, entirely up to you

PR #243, opened 2022-11-17
chore: expose more block members

ChaoticTempest:
This adds a few more members to the `Block` type since only the basic stuff was there before. These fields are useful if someones trying to monitor what's on the chain.


workspaces/src/types/block.rs
```diff
@@ -32,18 +37,42 @@ impl Block {
...
+    pub next_bp_hash: CryptoHash,
+    pub latest_protocol_version: u32,
+
+    pub prev_state_root: CryptoHash,
+    pub chunk_receipts_root: CryptoHash,
+    pub chunk_headers_root: CryptoHash,
+    pub chunk_tx_root: CryptoHash,
+    pub outcome_root: CryptoHash,
+    pub challenges_root: CryptoHash,
+    pub block_merkle_root: CryptoHash,
```
frol, line 67: Instead of making these fields public, I suggest we expose getters. That way it will be easier to rename/remove/add fields when we need it, without breaking changes.

frol (changes requested):

workspaces/src/types/block.rs
```diff
@@ -1,11 +1,16 @@
...
 
 /// Struct containing information on block coming from the network
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
 pub struct Block {
-    header: BlockHeader,
+    pub author: AccountId,
+    pub header: BlockHeader,
+    pub chunks: Vec<ChunkHeader>,
```
frol, line 13: @shariffdev Let's also update Block and ChunkHeader types to follow the approach of using getters instead of exposing internal fields (`#[non_exhaustive]` should be removed then)

frol (approved):

PR #245, opened 2022-11-19
fix: Make call consistent with worker::view

ChaoticTempest:
Since `Worker::view` got changed to be a builder, I noticed that `Worker::call` is no longer consistent with it. So best to make it also return a builder similar to `Account::call` as well. This should come in before [release/0.7.0](https://github.com/near/workspaces-rs/pull/244).

Also, exposed `{Account, Contract}::signer()` since that has been public and can be utilized with `Worker` related methods. Internally cleaned up the storage of `AccountId` in `Account` since `InMemorySigner` already holds the account id.


ChaoticTempest:
Merging this one in since there's only one API breakage, and the `signer` method being exposed. The rest is internal anyways

PR #246, opened 2022-12-01
fix changelog typo

austinabell:


ChaoticTempest (approved):
whoops, thanks for noticing

Issue #223, opened 2022-10-24
Fix patch state not being able to patch for large amounts of data


ChaoticTempest:
When patching a large payload to contract state, the sandbox near node would not it, throwing the following error:
```
Error: the server returned a non-OK (200) status code: [413 Payload Too Large]
```

This is probably to do with the fact that the network can be attacked if the requests are too big. But since this is sandbox, we can just special case it to accept large payloads.

PR #251, opened 2022-12-07
chore: configure sandbox

ChaoticTempest:
Separated out the logic purely from https://github.com/near/workspaces-rs/pull/241. This allows configuring `config.json` so that sandbox can do extra stuff on top of what mainnet/testnet does such as being able to patch large states which resolves https://github.com/near/workspaces-rs/issues/223


workspaces/src/network/config.rs
```diff
@@ -0,0 +1,156 @@
...
+
+use serde_json::{Map, Value};
+
+use crate::error::ErrorKind;
+use crate::Result;
+
+/// Replaces all entries recursively from `from` into `into`.
+fn overwrite_map(from: Map<String, Value>, mut into: Map<String, Value>) -> Result<Value> {
+    for (key, from_val) in from.into_iter() {
+        let Some(into_val) = into.remove(&key) else {
```
DavidM-D, line 25: This is unstable rust.
Do we require unstable to compile this crate?

workspaces/src/network/config.rs
```diff
@@ -0,0 +1,156 @@
...
+        // Default is 1GB which should suit most contract sizes.
+        Err(_err) => 1024 * 1024 * 1024,
+    };
+
+    Ok(max_files)
+}
+
+/// Set extra configs for the sandbox defined by workspaces.
+pub(crate) fn set_sandbox_configs(home_dir: &Path) -> Result<()> {
+    overwrite(
```
DavidM-D, line 95: Can this be replaced by [json_patch's merge](https://docs.rs/json-patch/0.2.7/json_patch/fn.merge.html) and a bit of IO?

DavidM-D (approved):
Logic looks good, but can we just replace most of this with this lib?

https://docs.rs/json-patch/0.2.7/json_patch/fn.merge.html

workspaces/src/network/config.rs
itegulov: Are you talking about the destructing assignment? This was stabilized a year ago I think

ChaoticTempest:
@DavidM-D didn't even know that library existed. At first was trying to find similar functionality inside `serde_json`

DavidM-D:
Not 100% sure we need to test `json_patch`, but your call and LGTM

ChaoticTempest:
I was keeping it just in case json_patch changes, but we're pinning the version anyways

PR #258, opened 2023-01-07
fix: storing credentials

ChaoticTempest:
Notice this minor issue where storing the credentials doesn't store the parent id for a subaccount (i.e. `sub.parent` would get stored as `sub.json`). This was due to `Path::set_extension("json")`  setting `parent` into `json` which I didn't notice earlier till now.

### DIFF
--- a/workspaces/src/rpc/tool.rs
+++ b/workspaces/src/rpc/tool.rs
@@ -57,15 +57,20 @@ pub(crate) async fn url_create_account(
     Ok(())
 }
 
-pub(crate) fn write_cred_to_file(path: &Path, id: &AccountId, sk: &SecretKey) {
-    let mut file = File::create(path).expect("Failed to create / write a key file.");
+pub(crate) fn write_cred_to_file(path: &Path, id: &AccountId, sk: &SecretKey) -> Result<()> {
+    let mut file = File::create(path).map_err(|err| {
+        ErrorKind::Io.full(
+            format!("failed to open {path:?} for writing credentials"),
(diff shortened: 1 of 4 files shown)


workspaces/src/types/account.rs
```diff
@@ -180,15 +180,10 @@ impl Account {
...
-
-        let mut savepath = savepath.join(self.id().to_string());
-        savepath.set_extension("json");
-
-        crate::rpc::tool::write_cred_to_file(&savepath, self.id(), &self.secret_key().0);
-
-        Ok(())
+        let savepath = save_dir.as_ref();
+        std::fs::create_dir_all(&save_dir).map_err(|e| ErrorKind::Io.custom(e))?;
+        let savepath = savepath.join(format!("{}.json", self.id()));
```
miraclx, line 185: Nice catch!

workspaces/src/rpc/tool.rs
```diff
@@ -85,7 +90,6 @@ pub(crate) fn write_cred_to_file(path: &Path, id: &AccountId, sk: &SecretKey) {
     .to_string()
     .into_bytes();
 
-    if let Err(err) = file.write_all(&content) {
-        panic!("Failed to write a key file {}", err);
-    }
+    file.write_all(&content)
+        .map_err(|err| ErrorKind::Io.full("Failed to write a key file", err))
```
miraclx, line 94: +1

miraclx (approved):

PR #259, opened 2023-01-09
feat: use tempdir crate instead of provided by os

aleksuss:
The PR helps to prevent clogging of OS temp directory.


ChaoticTempest:
hey thanks for the PR, but `tempfile` was added already with the [port collision fix](https://github.com/near/workspaces-rs/pull/257). So I'll close this one unless you think there's something missing from it

Issue #263, opened 2023-01-17
`workspaces::types::PublicKey` doesn’t implement `std::hash::Hash`


mooori:
This prevents storing it in a `std::collections::HashSet`, which can be helpful during testing.

`workspaces::types::PublicKey` is a [wrapper](https://docs.rs/workspaces/latest/src/workspaces/types/mod.rs.html#76) around `near_crypto::PublicKey`, which [implements](https://docs.rs/near-crypto/0.15.0/near_crypto/enum.PublicKey.html#impl-Hash-for-PublicKey) `Hash`. So it should be possible for `workspaces::types::PublicKey` to derive `Hash`. Would a PR that adds `Hash`to the derived traits be welcome?

ChaoticTempest:
Oh thanks for bringing this up. Lots of types in workspaces are there more on a need basis, so some stuff might've been missed. There's a couple things missing from `PublicKey` that would be nice to have as well. I'll include this in that PR as well

PR #265, opened 2023-01-19
chore: Import some functions over from near_crypto for PublicKey

ChaoticTempest:
This just adds a couple convenience methods over from `near_crypto` crate. This was originally requested with just `Hash` impl in https://github.com/near/workspaces-rs/issues/263, but decided to import a couple more functions as they are convenient to have if people are managing public keys with workspaces

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - [Import a couple functions over from near_crypto for PublicKey](https://github.com/near/workspaces-rs/pull/265)
-  - Impl `Ord`, `PartialOrd`, `Hash`, `BorshSerialize`, `BorshDeserialize`, `Display`, `FromStr`, `TryFrom<&[u8]>`, and `Into<Vec<u8>>` for `PublicKey`
+  - Impl `Ord`, `PartialOrd`, `Hash`, `BorshSerialize`, `BorshDeserialize`, `Display`, and `FromStr` for `PublicKey`
   - Added `PublicKey::{empty, len, key_data}`
   - Impl `Display` for `SecretKey`.
   - more docs were added to both `SecretKey` and `PublicKey`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Added
 
+- [Import a couple functions over from near_crypto for PublicKey](https://github.com/near/workspaces-rs/pull/265)
+  - Impl `Ord`, `PartialOrd`, `Hash`, `BorshSerialize`, `BorshDeserialize`, `Display`, `FromStr` for `PublicKey`
+  - Added `PublicKey::{empty, len, key_data}`
+  - Impl `Display` for `SecretKey`.
+  - more docs were added to both `SecretKey` and `PublicKey`.
+
 ### Changed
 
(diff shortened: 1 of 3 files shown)


workspaces/src/types/mod.rs
```diff
@@ -72,13 +105,82 @@ impl From<PublicKey> for near_crypto::PublicKey {
...
+    }
+}
+
+impl From<PublicKey> for Vec<u8> {
+    fn from(pk: PublicKey) -> Vec<u8> {
+        let mut data = vec![pk.key_type() as u8];
+        data.extend(pk.key_data());
+        data
+    }
+}
```
miraclx, line 173: This sort of feels like an antipattern, let's maybe just leave it to borsh?

ChaoticTempest: you have a good point there 👍 . Originally ported this over from SDK, but the borsh call is much more explicit anyways since it's called `try_to_vec` instead of purely just `into`

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 ### Added
 
+- [Import a couple functions over from near_crypto for PublicKey](https://github.com/near/workspaces-rs/pull/265)
+  - Impl `Ord`, `PartialOrd`, `Hash`, `BorshSerialize`, `BorshDeserialize`, `Display`, `FromStr`, `TryFrom<&[u8]>`, and `Into<Vec<u8>>` for `PublicKey`
```
miraclx, line 8:
```suggestion
  - Impl `Ord`, `PartialOrd`, `Hash`, `BorshSerialize`, `BorshDeserialize`, `Display`, and `FromStr` for `PublicKey`
```

workspaces/tests/types.rs
```diff
@@ -33,6 +35,58 @@ fn test_keypair_secp256k1() -> anyhow::Result<()> {
...
+
+        // invalid public key should error out on deserialization:
+        assert!(PublicKey::try_from_slice(&[0]).is_err());
+    }
+
+    Ok(())
+}
+
+#[test]
+fn test_pubkey_borsh_format_change() -> anyhow::Result<()> {
```
miraclx, line 56: Hm. Struggling a bit to understand this test.

Is there a case where this would fail that's directly impacted by the struct here? Otherwise, it looks like a data test is all we're trying to do.

So this would be equivalent, no?

```rust
let mut data = vec![KeyType::ED25519 as u8];
data.extend(bs58::decode("6E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp").into_vec()?);

let new_key = PublicKey::try_from_slice(data.as_slice())?;

assert_eq!(
    new_key.try_to_vec()?,
    bs58::decode("16E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp").into_vec()?
);
```

ChaoticTempest: hmm, now that you mention it, yeah it should be equivalent. I wonder why SDK did it like this in the first place

ChaoticTempest: oh it's because it had `Vec<u8>.try_into()` but I just got rid of it in favor of borsh

miraclx: Ah, gotcha!

miraclx (approved):
Lgtm!

Issue #262, opened 2023-01-17
Convert `near_sdk::PublicKey` to `workspaces::types::PublicKey`


mooori:
Some tests need to compare `PublicKey`s from the two crates mentioned above ([example](https://github.com/aurora-is-near/near-plugins/blob/33b4a2cff630be2ebe88e3301b7caf7361c4d6dc/near-plugins/tests/full_access_key_fallback.rs#L150-L171)).

It seems like there’s no straightforward way for the conversion since the only way to generate a `workspaces::types::PublicKey` from data is [Deserialize](https://docs.rs/workspaces/0.7.0/workspaces/types/struct.SecretKey.html#impl-Deserialize%3C%27de%3E-for-SecretKey). Which leads to [this](https://github.com/aurora-is-near/near-plugins/blob/33b4a2cff630be2ebe88e3301b7caf7361c4d6dc/near-plugins/tests/full_access_key_fallback.rs#L28-L37) inefficient and awkward conversion.

Maybe there’s a more efficient and elegant conversion which I’ve been missing? If not, it would be great if such a conversion could be supported.

`workspaces::types::PublicKey` is [just a wrapper](https://docs.rs/workspaces/0.7.0/src/workspaces/types/mod.rs.html#76) around `near-crypto::PublicKey`. So for `workspaces` it might be possible to support the conversion via:

- Get the bytes of a `near_sdk::PublicKey` ([as_bytes()](https://docs.rs/near-sdk/4.1.1/near_sdk/struct.PublicKey.html#method.as_bytes)).
- Convert the bytes to `near_crypto::PublicKey` (e.g. via [this function](https://docs.rs/near-crypto/0.15.0/near_crypto/struct.ED25519PublicKey.html#method.try_from)) and then wrap it in `workspaces::types::PublicKey`.

PR #267, opened 2023-01-23
chore: add `sdk::PublicKey` to `workspaces::PublicKey` conversion

ChaoticTempest:
Resolves https://github.com/near/workspaces-rs/issues/262

This adds a `TryFrom<sdk::PublicKey>` impl to convert `sdk::PublicKey` into `workspaces::PublicKey`s, along with a test to check serialization. This also has a new feature flag `interop_sdk` just in a case they don't want to add a big dependency to their testing pipeline. It's enabled by default anyways

Borsh serialization had to change from the previous PR https://github.com/near/workspaces-rs/pull/265 to make it compatible to be sent to a contract through borsh serialization. The new format is now `[buffer_len, key_type, key_data..]`


miraclx (commented):
Just took a look at `sdk::PublicKey`'s borsh-serialization format, and it's really weird tbh, I think in general it's not ideal to assume byte slices deserialize the same way for two types in differing crates with the same name. We should be able to retain the current behavior, and instead just provide two-way conversion at the API-level.

So, if you have a byte slice that successfully deserializes into `sdk::PublicKey`, you can just call `into()` on that instead of attempting to deserialize the same byte slice into a `workspaces::PublicKey`. And vice-versa.

workspaces/src/types/mod.rs
```diff
@@ -65,6 +69,15 @@ impl KeyType {
             near_crypto::KeyType::SECP256K1 => Self::SECP256K1,
         }
     }
+
+    /// Length of the bytes of the public key associated with this key type.
+    #[allow(clippy::len_without_is_empty)] // This is an associated length.
+    pub const fn len(&self) -> usize {
```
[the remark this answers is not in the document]
ChaoticTempest: No clear public use cases based on data or anything, but just seemed nice to have when people just have the `KeyType` accessible instead of the `PublicKey` for whatever reason

ChaoticTempest:
> Just took a look at sdk::PublicKey's borsh-serialization format, and it's really weird tbh, I think in general it's not ideal to assume byte slices deserialize the same way for two types in differing crates with the same name. We should be able to retain the current behavior, and instead just provide two-way conversion at the API-level.

So I was trying to eliminate a friction point with this, but historically for serialized types in workspaces, it's just been passing them into `args*` functions without any sort of conversions into sdk types like so:
```rust
contract.call("function")
    .args_json(json!({
         "account_id": AccountId("user.test"),
    })
    .transact().await?;
```

Adding the conversion would feel pretty inconsistent:

```rust
contract.call("function")
    .args_json(json!({
         "account_id": AccountId("user.test"),
         "public_key": sdk::PublicKey::from(PublicKey::try_from_bytes(&bytes)?)
    })
    .transact().await?;
```
since we'd now need to know what types are serializable via workspaces or sdk. I'd like not to go down that route, since the experience wouldn't be nice I'd say.

> So, if you have a byte slice that successfully deserializes into sdk::PublicKey, you can just call into() on that instead of attempting to deserialize the same byte slice into a workspaces::PublicKey. And vice-versa.

Like creating `sdk::PublicKey` won't have an issue if you have the byte slice, but it's a separate issue when workspaces would give you `workspaces::PublicKey`s from its APIs, which isn't ideal cause of the inconsistency with serializing them later mentioned above.

workspaces/src/types/mod.rs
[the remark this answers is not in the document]
miraclx: Okay, although I'm concerned `len` might be a misnomer, considering rust's naming convention. Maybe we can name it after it's equivalent in sdk? https://github.com/near/near-sdk-rs/blob/89f8c2d776c8d00feb45239fdb15725527f1894b/near-sdk/src/types/public_key.rs#L23

miraclx:
> but it's a separate issue when workspaces would give you workspaces::PublicKeys from its APIs, which isn't ideal cause of the inconsistency with serializing them later mentioned above.

That's a fair point, although just noting that serde-serialization looks to be consistent between sdk/workspaces' `PublicKey`s. But I understand the inclination to implicitly pass one type for another, especially since there's no strictly typed checks in that context. Inconsistent borsh formats moves into a runtime issue. Could definitely save a lot of hair pulling.

workspaces/src/types/mod.rs
```diff
@@ -128,6 +129,19 @@ impl PublicKey {
         Self(near_crypto::PublicKey::empty(key_type.into_near_keytype()))
     }
 
+    /// Create a new [`PublicKey`] from the given bytes. This will return an error if the bytes are not in the
+    /// correct format. Expected to have key type be the first byte encoded, with the remaining bytes being the
+    /// key data.
+    pub fn try_from_bytes(bytes: &[u8]) -> Result<Self> {
```
miraclx, line 135: I think we should reduce the scope of exposure here, if there's no way to generate a slice from `workspaces::PublicKey` that this method can then parse, we probably shouldn't have this method public.

```suggestion
    pub(crate) fn try_from_bytes(bytes: &[u8]) -> Result<Self> {
```

workspaces/src/types/sdk.rs
```diff
@@ -0,0 +1,18 @@
+use std::convert::TryFrom;
+
+use crate::error::{Error, ErrorKind};
+
+use super::PublicKey;
+
+impl TryFrom<near_sdk::PublicKey> for PublicKey {
```
miraclx, line 7: PS: what do you think about providing an inverse conversion?

workspaces/src/types/mod.rs
ChaoticTempest: I'm on the fence with keeping or not keeping it, but I was thinking we'd have a raw bytes API still so that people could construct it if they have the info from a different source than just workspaces or sdk. `BorshDeserialize::try_from_slice` would probably not be intuitive since we can't attach any docs noting that it requires 4 extra bytes denoting the length at the start.

workspaces/src/types/sdk.rs
ChaoticTempest: what would the use case be for the inverse? I don't think anything would consume sdk related types besides sending it over the wire through serialization

workspaces/src/types/mod.rs
miraclx: > but I was thinking we'd have a raw bytes API still so that people could construct it if they have the info from a different source than just workspaces or sdk.

So the input here is the borsh-serialized format of `near-crypto::PublicKey`.

So they either already have a `near_crypto::PublicKey` or they manually constructed the slice.

- For the first, perhaps we just add a `From<near_crypto::PublicKey>` for `workspaces::PublicKey`?
- For the second, this signature could be `pub fn from_parts(key_type: KeyType, key_data: &[u8]) -> Result<Self>`

What do you think?

workspaces/src/types/sdk.rs
miraclx: That makes sense. Didn't have any specific use case in mind. Just thought it made sense to provide all the reverse conversions where possible. But if there's no clear use case for this, it's fine to leave out.

workspaces/src/types/mod.rs
frol: Why do we need a copy-pasted implementation of `KeyType` in the first place? Wouldn't `near_crypto` types work here? (we already have it as a dependency)

ChaoticTempest: it's because we don't want to expose it as a transient dependency quiet yet, since that would lead us to maintaining that relationship going forward and cause unwanted breaking changes along the way if `near_crypto` were to change. So, until `near_crypto` is stable (i.e. 1.0), then we shouldn't expose it and just maintain a version ourselves to have this level of control. This was also done for things like `near_primitives` as well to avoid that hassle

frol: @ChaoticTempest Thanks for the context!

ChaoticTempest:
@frol this should be good to go whenever you want to fully take a look at it

frol (approved):
I am not exactly sure about having `interop_sdk` to be enabled by default, but I think in the bright future where PublicKey is extracted into a separate crate and shared across nearcore, near-sdk-rs, and workspace-rs, we can drop this code altogether and the feature will become no-op

ChaoticTempest:
@frol by enabling it by default, it offers the least friction to testing contracts which `workspaces` should mostly be geared towards. It shouldn't be too huge of a concern since this is during testing time. Where it becomes bloatware is when `workspaces` gets used for automation or as a near-api-rs replacement, but I don't think that's a huge concern either since people can disable the feature if it becomes too bothersome.

> [...] but I think in the bright future where PublicKey is extracted into a separate crate and shared across nearcore, near-sdk-rs, and workspace-rs, we can drop this code altogether and the feature will become no-op

So, the `PublicKey` in `near-sdk-rs` isn't exactly the same as the one used in `near-primitives`. So there's the extra requirement that `PublicKey` is the same and consistent between all three libraries. I fear the `near_sdk::PublicKey` would be harder to make consistent with the `near_primitives` one as that one has stricter requirements like being wasm-compilable. The alternative I went for here is to just keep the byte format the same between the two

PR #269, opened 2023-03-07
Bump up versions of the dependencies.

silvestrpredko:


silvestrpredko:
Unfortunately, seems like my PR brokes everything.

ChaoticTempest:
@silvestrpredko  It looks like you deleted the workspaces types which was hiding `CryptoHash` before. It's not the same interface so things likely broke internally. We're waiting on `near_primitives` to be in 1.0 version before we reliably use it directly

silvestrpredko:
Yeah seems like you are right. By the way, I am working on a library client for Near. https://github.com/Relayz-io/near-client It's basically an implementation of a NearRPC. And I am dependent on near-primitives. It would be great to see the stable version soon.

ChaoticTempest:
I see, yeah let's just bump up the versions without deleting that workspaces types then. There's no timeline for the stable version of near-primitives. 0.16 version shouldn't conflict besides needing to bump the Minimum support rust version to 1.67.1

silvestrpredko:
@ChaoticTempest Done ✅. Please check it out.

ChaoticTempest:
closing this in favor of https://github.com/near/workspaces-rs/pull/271 since the tests are passing for those

PR #274, opened 2023-05-15
feat: support manually supplied validator key

### LINKED ISSUE #273: `Worker<Sandbox>` presumes the process is running in the same filesystem as sandbox
#220 presumes that the sandbox instance is running on the same filesystem as the current process. Specifically, [this](https://github.com/near/workspaces-rs/blob/main/workspaces/src/network/sandbox.rs#L36-L39) tries to pull the validator keys from the local filesystem, while the sandbox process might be running in a virtualized container or remotely.

We could either provide a configuration parameter for `NetworkBuilder` presuming that the user can fetch the keys themselves or add a sandbox endpoint that returns the keys in plaintext. Not sure if there is anything easier/more convenient.

itegulov:
Fixes #273

### DIFF
--- a/workspaces/src/network/builder.rs
+++ b/workspaces/src/network/builder.rs
@@ -1,9 +1,10 @@
 use std::future::{Future, IntoFuture};
 use std::marker::PhantomData;
-use std::path::PathBuf;
 
 use crate::network::Sandbox;
 use crate::{Network, Worker};
+
+use super::server::ValidatorKeyTactic;
 
 pub(crate) type BoxFuture<'a, T> = std::pin::Pin<Box<dyn Future<Output = T> + Send + 'a>>;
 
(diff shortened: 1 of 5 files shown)


ChaoticTempest (approved):

workspaces/src/network/builder.rs
```diff
@@ -1,10 +1,11 @@
 use std::future::{Future, IntoFuture};
 use std::marker::PhantomData;
-use std::path::PathBuf;
 
 use crate::network::Sandbox;
 use crate::{Network, Worker};
 
+use super::server::ValidatorKeyTactic;
```
ChaoticTempest, line 7: actually, how about we just call it `ValidatorKey`. The `Tactic` part doesn't add much anyways

ChaoticTempest:
@itegulov good to merge?

PR #275, opened 2023-06-17
Enhancement : Use cargo near to build project

### LINKED ISSUE #230: Replace ad-hoc build logic with `cargo-near`

ghimire007:
replaced existing build logic to use cargo near to resolve #230

### DIFF
--- a/workspaces/src/cargo/mod.rs
+++ b/workspaces/src/cargo/mod.rs
@@ -8,9 +8,17 @@ use std::fs;
 use std::io;
 use std::path::Path;
 use std::process::Stdio;
+use std::convert::TryInto;
 
 use crate::error::ErrorKind;
 use crate::Result;
+
+
+use cargo_near::build::run;
+use cargo_near::{BuildCommand, ColorPreference};
+use cargo_near::util::CompilationArtifact;
+
+
 
 fn cargo_bin() -> Command {
     match env::var_os("CARGO") {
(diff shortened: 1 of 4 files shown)


frol:
@ghimire007 I [cleaned up](https://github.com/near/workspaces-rs/pull/275/commits/def91b1c82ad77c111b531b712c3a1aac8244707) the code even further. This PR is only waiting for a new cargo-near release (crates with `git` dependencies cannot be published to crates.io), which is blocked on https://github.com/near/cargo-near/pull/97#issuecomment-1597789772

ghimire007:
@frol I have updated project  to use cargo-near 0.3.1

workspaces/src/cargo/mod.rs
```diff
@@ -1,129 +1,41 @@
...
-            "Compilation resulted in more than one '.wasm' target file: {:?}",
-            wasm_files
-        ))),
-    }
+    let cargo_near_build_command = cargo_near::BuildCommand {
+        release: true,
+        embed_abi: true,
+        doc: false,
+        color: cargo_near::ColorPreference::Always,
+        no_abi: true,
```
frol, line 22: Just for future context, there was no analysis on which flags actually make sense to use, so we just followed the existing setup - a simple release-mode build for wasm target.

frol (approved):

Issue #276, opened 2023-06-21
Prevent running server on exposed network address


agostbiro:
When I run the workspace as part of an integration test I get this popup on the latest MacOS:

![Screenshot 2023-06-21 at 11 39 45](https://github.com/near/workspaces-rs/assets/5764438/a32cbf29-f69f-4102-b8d1-e8195572c526)

This indicates that the server tries to listen on an exposed network address. Ideally, the server would only run on localhost by default and then this popup wouldn't show.

A quick look at the [source](https://github.com/near/workspaces-rs/blob/26ca5aee8477fcf59fdb25bfa4a612989864e4da/workspaces/src/network/server.rs#LL18C36-L18C36) shows that the default address is indeed localhost:

```rust
pub const DEFAULT_RPC_URL: &str = "http://localhost";
```

so there must be something weird going on to get this popup. 

Happy to dig in and provide a PR if you think it's important.

frol:
@agostbiro Good catch. Yeah, it is indeed annoying and it is better to solve it. I believe it is not RPC that is not running on localhost, but the P2P part of neard. Workspaces-rs sandbox launches neard server and neard uses two TCP ports: (1) [RPC](https://github.com/near/nearcore/blob/140c573a8fee995f0590fda52a965a91fe31bcbf/nearcore/src/config.rs#L287), (2) [p2p network](https://github.com/near/nearcore/blob/140c573a8fee995f0590fda52a965a91fe31bcbf/nearcore/src/config.rs#L292). Workspaces-rs patches the config to specify the proper RPC IP:address:

https://github.com/near/workspaces-rs/blob/c31a955df8a55a6c82cedaaf820ed9236ee1fee5/workspaces/src/network/server.rs#L100

My guess is that it does not patch the `network` addr (I see that it seems to patch the port)

Please, address it.

agostbiro:
Thanks @frol , that was helpful! It turns out that there were two separate causes for the popups, so it was a bit involved to track them down, but the fix is in now: https://github.com/near/workspaces-rs/pull/277

frol:
Resolved in #277

PR #277, opened 2023-06-27
Run `neard` on `localhost` instead of `0.0.0.0` to prevent firewall popups on MacOS

agostbiro:
This PR fixes the problem where users were getting firewall popups when running integration tests on MacOS as described in #276.

It turns out this had two causes:

1. `neard` was running on `0.0.0.0` instead of `localhost`.
2. The [`portpicker`](https://github.com/Dentosal/portpicker-rs) library used to acquire free random ports was testing ports by binding to `0.0.0.0`.

This PR fixes both causes. 

The PR changes how `neard` is ran by using the more verbose `run_with_options` where we can specify CLI arguments to `neard`. We use this to specify the IP address as localhost. Previously only the port numbers were specified and `neard` defaulted to running on `0.0.0.0`.

This PR also removes the `portpicker` library and provides a simple function to pick random port numbers with the OS feature where port number 0 is replaced by the OS with a random unused port. I initially considered contributing a fix to the `portpicker` library, but it seems unmaintained, so I thought it was a better choice to remove it from dependencies.

There are two differences between the implementation in the PR and the `portpicker` library:

1. The `portpicker` library also checks if the port returned by the OS is a valid UDP port. The implementation in this PR forgoes that check as I think we only need TCP ports for `neard`.
2. The `portpicker` library tries to acquire a random port by testing random numbers before falling back to requesting a port with port 0 from the OS. It's unclear why it does this. It might be a mechanism in case the OS doesn't support requesting random ports with port 0. All modern UNIX operating systems and Windows should support this, so I don't think we need such a fallback mechanism, but let me know if we do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 ### Fixed
 
+- [Run `neard` on `localhost` instead of `0.0.0.0` to prevent firewall popups on MacOS](https://github.com/near/workspaces-rs/issues/276)
+
 ## [0.7.0]
 
 ### Added
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -20,7 +20,6 @@ cargo-near = "0.3.1"
 chrono = "0.4.19"
 fs2 = "0.4"
 hex = "0.4.2"
-portpicker = "0.1.1"
 rand = "0.8.4"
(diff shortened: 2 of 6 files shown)


frol (approved):
@agostbiro Looks good to me! Let's get it finally fixed

PR #280, opened 2023-07-06
fix: fix typos

itsyaasir:
As I was going through the code source, I have some small typos here and there, this is small PR aims to fix that.


itsyaasir:
@frol A small PR to fix typos in the source code.

frol (approved):
@itsyaasir Thanks for fixing those! Sorry for the delay, I got really underwater with several weeks of offline events back-to-back. I am now back to getting to normal operations.

Issue #29, opened 2021-12-03
Add API for measuring gas


ChaoticTempest:
Currently, we'd comb through the `CallExecution` outcomes from certain functions that emit this value such as `call` (but not all function will emit it) to get the used gas. And then later, we add it all up to get the total gas consumed. We can ease this by having some sort of `GasMeter` that would measure changes between each call, or whenever we want to get the elapsed gas used. 

This would potentially look like this:
 ```rust
 let meter = GasMeter::now(worker);
 ...
 worker.call(...);
 ...
 println!("Gas used: {} yoctoNear", meter.elapsed().as_yocto());

```

**Comment by itsyaasir** [comment]
@frol This issue looks interesting, while I got the general idea of the issue, can someone expound it more? Thanks

In the meantime am still studying up the source code.

**Comment by itsyaasir** [comment]
- One idea I got for solving this is, after we initialize the `GasMeter`, and do bunch of calls, we can comb through the blocks to get the transaction outcomes and add up the gas used. (I feel this is inefficient)
- It will be helpful if someone could guide me through this as am new to the nearprotocol.

Any other suggestions are welcome, Thanks.

**Comment by ChaoticTempest** [comment]
When I was writing the reference syntax, what I had in mind was to have `GasMeter` directly hook onto when worker would make `call`s and add up all the gas it's seen. Worker here shouldn't be storing anything but should offer hooks like `set_on_transact(fn)`. This is a bit more difficult ticket if you're unfamiliar with the codebase as it requires restructuring it to have these hooks, but feel free to play around with the idea

**Comment by frol** [comment]
@ChaoticTempest Do I understand correctly that you want to measure only the gas used specifically for the function call execution and ignore all the other costs? What is the benefit over just using the total gas used?

If I understood this issue correctly, #284 is not going in the right direction, right? I think it would be helpful if you share the problem you think we can solve with this feature.

**Comment by ChaoticTempest** [comment]
@frol this feature is meant to allow monitoring gas usage over multiple calls/transactions, so people don't have to manually add them all up themselves.

Roughly, this is what it should do:
```rust
let meter = GasMeter::now(&worker);
account.call(...)
acocunt.call(...)
...
let total_gas_usage = meter.elapsed();
println!("{}", total_gas_usage);
```

This is purely syntax sugar over:
```rust
let mut total_gas_usage = 0;
let result = account.call(...)
total_gas_usage += result.total_gas_burnt;

let result = account.call(...)
[...]

PR #284, opened 2023-08-01
feat: gas hook on ExecutionFinalResult

shariffdev:
This adds an `on_transaction` hook to the ExecutionFinalResult


shariffdev:
@frol Since I can't invite reviewers currently, any comments from reviewers are welcome.

frol (changes requested):
@shariffdev I don't think it is what was described in the issue, see my comment there: https://github.com/near/workspaces-rs/issues/29#issuecomment-1674985992

shariffdev:
Checking right now

> @shariffdev I don't think it is what was described in the issue, see my comment there: [#29 (comment)](https://github.com/near/workspaces-rs/issues/29#issuecomment-1674985992)

ChaoticTempest:
From what I see, this is only hooking onto the current transaction. We want to have the callback be executed on all transactions since the creation of the `GasMeter`. `Worker` here should still store a list of hooks so that it can call back on all of them whenever it receives a transaction result from the network. Sorry if my previous wording there didn't make sense.

What I suggest is to have a very similar API to Rust's [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) so that when it's created via `GasMeter::now(&worker)`, it doesn't need to know about previous calls similar to the `Instant::now` API for time.

shariffdev:
Hi @ChaoticTempest help you check this

frol:
@shariffdev Please, make the CI pass. I see there is a dependency issue, use the following workaround: https://github.com/near/near-sdk-rs/pull/1075/files#diff-748c9b7c0e525461124dec4c258406a92d81a9b610f7b47a79efe9a0cb52553c

shariffdev:
@frol The error is related to the clap requiring a newer rust version (1.70) I am looking into creating a new PR which we can update rustc to 1.70+, This [issue](https://github.com/near/nearcore/issues/9143) seems to have been resolved so we can update to the latest 1.70+

shariffdev:
This [PR](https://github.com/near/workspaces-rs/pull/293) should take care of the dependency issue as we are using a newer stable toolchain

workspaces/src/worker/mod.rs
```diff
@@ -13,6 +14,7 @@ use crate::{Network, Result};
 /// deploying a contract, or interacting with transactions.
 pub struct Worker<T: ?Sized> {
     pub(crate) workspace: Arc<T>,
+    pub(crate) on_transact: Option<GasHook>,
```
ChaoticTempest, line 17: This is only creating a singular gas hook, and would overwrite the previous hook if multiple `GasMeters` were created. Best to have vector of gas hooks instead

workspaces/src/types/gas_meter.rs
```diff
@@ -0,0 +1,72 @@
...
+///    .call("set_status")
+///    .args_json(json!({
+///        "message": "hello_world",
+///    }))
+///    .transact()
+///    .await?;
+///
+/// println!("Total Gas consumed: {}", meter.elapsed()?);
+/// ```
+pub struct GasMeter {
```
ChaoticTempest, line 33: What about the case where `GasMeter` is dropped? `Worker` now would be calling into gas hooks that are no longer alive or necessary

workspaces/src/types/gas_meter.rs
```diff
@@ -0,0 +1,72 @@
...
+        let mut meter = self.gas.lock()?;
+        *meter = 0;
+        Ok(())
+    }
+
+    /// Add gas to the total amount of gas consumed.
+    pub fn add_gas(&self, g: Gas) {
+        let mut locked_gas = self.gas.lock().expect("GasMeter lock poisoned");
+        *locked_gas += g;
+    }
```
ChaoticTempest, line 71: This function doesn't seem to be used anywhere. Is it necessary?

shariffdev: No, it is not necessary!

workspaces/src/types/gas_meter.rs
shariffdev: I found channels work well for our use case instead of passing hooks. Acceptable?

workspaces/src/types/gas_meter.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use crate::Worker;
+
+use super::Gas;
+use crate::result::Result;
+
+/// A hook that is called on every transaction that is sent to the network.
+/// This is useful for debugging purposes, or for tracking the amount of gas
+/// that is being used.
+pub type GasHook = Arc<Mutex<dyn FnMut(Gas) -> Result<()> + Send>>;
```
ChaoticTempest, line 14: we actually don't need to have a mutex here. The function never gets mutated, so this isn't necessary anymore.
```suggestion
pub type GasHook = Arc<dyn Fn(Gas) -> Result<()> + Send + Sync>;
```

workspaces/src/worker/mod.rs
```diff
@@ -13,6 +14,7 @@ use crate::{Network, Result};
 /// deploying a contract, or interacting with transactions.
 pub struct Worker<T: ?Sized> {
     pub(crate) workspace: Arc<T>,
+    pub(crate) on_transact: Vec<GasHook>,
```
ChaoticTempest, line 17: This is now a vector of hooks, so we should rename it to be plural, something like `tx_callbacks`

workspaces/src/types/gas_meter.rs
```diff
@@ -0,0 +1,76 @@
...
+        worker
+            .on_transact
+            .push(Arc::new(Mutex::new(move |gas: Gas| {
+                // check if reference is unique, expect false
+                match Arc::get_mut(&mut gas_consumed) {
+                    Some(_) => tracing::info!("GasMeter dropped, hook invalid."),
+                    None => gas_consumed.lock()?.add_assign(gas),
+                }
+
+                Ok(())
```
ChaoticTempest, line 58: The arc reference should not need to be checked like this, so we can just simplify to this:
```suggestion
                gas_consumed.lock()?.add_assign(gas);
                Ok(())
```

shariffdev: I wanted to add visibility that the meter is dropped. I made this alternative change:
```rust
 let gas_consumed = Arc::downgrade(&Arc::clone(&meter.gas));
 worker.tx_callbacks.push(Arc::new(move |gas: Gas| {
       // upgrades if meter is still alive, else noop.
      _ = gas_consumed.upgrade().map(|consumed| {
          consumed.lock().expect("meter is valid").add_assign(gas);
      });

      Ok(())
 }));
```

workspaces/src/operations.rs
```diff
@@ -236,13 +236,28 @@ impl Transaction {
...
+        if !self.worker.tx_callbacks.is_empty() {
+            let total_gas_burnt = view.transaction_outcome.outcome.gas_burnt
+                + view
+                    .receipts_outcome
+                    .iter()
+                    .map(|t| t.outcome.gas_burnt)
+                    .sum::<u64>();
+
+            self.worker.tx_callbacks.iter().for_each(|meter| {
+                meter(Gas::from_gas(total_gas_burnt)).unwrap();
```
ChaoticTempest, line 256: we shouldn't be calling into `unwrap` into all these meter functions. You already have a impl to convert PoisonError into an Error, so why not just return it with `?`

workspaces/src/types/gas_meter.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn now<T: ?Sized>(worker: &mut Worker<T>) -> Self {
+        let meter = Self {
+            gas: Arc::new(Mutex::new(Gas::from_gas(0))),
+        };
+
+        let gas_consumed = Arc::downgrade(&Arc::clone(&meter.gas));
+        worker.tx_callbacks.push(Arc::new(move |gas: Gas| {
+            // upgrades if meter is still alive, else noop.
+            _ = gas_consumed.upgrade().map(|consumed| {
+                let mut consumed = consumed.lock().expect("GasMeter lock poisoned");
```
ChaoticTempest, line 47: Same here. if additional context is needed, just map the error to a `Error.message`

ChaoticTempest (commented):
Aside from the error unwrappings, this one is good to go

workspaces/src/operations.rs
shariffdev: Done

workspaces/src/types/gas_meter.rs
shariffdev: Done

frol (approved):
@shariffdev Thanks for working on it and adding the documentation with an actionable example!

Issue #217, opened 2022-10-11
Expose `EXPERIMENTAL_*` RPC APIs


ChaoticTempest:
This might be nice for people to make use of and play around with these experimental features from workspaces side, but since these are `EXPERIMENTAL`, they should be constrained under the unstable feature flag or a separate experimental one

PR #285, opened 2023-08-03
feat: expose experimental apis

shariffdev:
Closes https://github.com/near/workspaces-rs/issues/217

The PR exposes experimental apis under an `experimental` flag.


workspaces/src/network/sandbox.rs
```diff
@@ -206,3 +234,106 @@ impl Sandbox {
...
+            .query_nolog(&RpcCheckTxRequest { signed_transaction })
+            .await
+            .map_err(|e| SandboxErrorCode::CheckTxFailure.custom(e))?;
+
+        Ok(resp)
+    }
+
+    // This method is a stub and does nothing.
+    pub(crate) async fn genesis_config(&self) -> Result<()> {
+        // NOTE: RpcGenesisConfigResponse is an empty struct with no fields, so don't do anything with it.
```
frol, line 282: Hmm, this does not sound true: https://docs.rs/near-jsonrpc-client/latest/near_jsonrpc_client/methods/EXPERIMENTAL_protocol_config/type.RpcProtocolConfigResponse.html it is clearly not an empty struct.

workspaces/src/network/sandbox.rs
```diff
@@ -206,3 +234,106 @@ impl Sandbox {
         Ok(())
     }
 }
+
+#[cfg(feature = "experimental")]
+impl Sandbox {
+    pub(crate) async fn changes_in_block(&self, block_reference: BlockReference) -> Result<()> {
+        // NOTE: RpcStateChangesInBlockByTypeResponse is an empty struct with no fields, so don't do anything with it.
```
frol, line 241: This is not true: https://docs.rs/near-jsonrpc-client/latest/near_jsonrpc_client/methods/EXPERIMENTAL_changes_in_block/struct.RpcStateChangesInBlockByTypeResponse.html

workspaces/src/network/sandbox.rs
```diff
@@ -206,3 +234,106 @@ impl Sandbox {
         Ok(())
     }
 }
+
+#[cfg(feature = "experimental")]
+impl Sandbox {
+    pub(crate) async fn changes_in_block(&self, block_reference: BlockReference) -> Result<()> {
```
frol, line 240: This is not the right place to implement RPC methods. RPC methods are not methods of "sandbox" they are methods of RPC client, see https://github.com/near/workspaces-rs/blob/cda1dc4e481e91bb17d95053b42d3b584cf2a95b/workspaces/src/rpc/client.rs, where other RPC methods are already implemented.

frol (changes requested):

workspaces/src/network/sandbox.rs
shariffdev: Missed that, I'll update

workspaces/src/network/sandbox.rs
shariffdev: Missed that, I'll update

workspaces/src/network/sandbox.rs
shariffdev: Thanks, I'll move the implementations

itsyaasir:
@frol This ready to be re-reviewed

shariffdev:

workspaces/src/rpc/client.rs
```diff
@@ -332,6 +350,147 @@ impl Client {
...
+        Ok(resp)
+    }
+
+    pub(crate) async fn signed_transaction<U: serde::Serialize>(
+        &self,
+        contract_id: &AccountId,
+        signer: &InMemorySigner,
+        func_name: String,
+        func_args: U,
+    ) -> Result<SignedTransaction> {
```
frol, line 466: Where did this method come from? This looks like `send_tx_and_retry` to me (it is already implemented above, so I don't see any reason for this method to exist)

workspaces/src/worker/impls.rs
```diff
@@ -173,6 +189,77 @@ where
     }
 }
 
+#[cfg(feature = "experimental")]
+impl<T: ?Sized> Worker<T>
+where
+    T: NetworkClient,
+{
+    pub async fn changes_in_block(
```
frol, line 197: @ChaoticTempest Why are JSON RPC methods are exposed through "worker"? This looks like an abstraction leak to me.

frol (changes requested):

examples/src/tx_status.rs
```diff
@@ -0,0 +1,35 @@
...
+        }))
+        .transact()
+        .await?;
+
+    println!("set_status: {:?}", outcome);
+
+    let tx_info = {
+        let outcome = outcome.outcome();
+        TransactionInfo::TransactionId {
+            hash: near_primitives::hash::CryptoHash(outcome.block_hash.0),
```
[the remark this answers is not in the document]
shariffdev: Would it be ok to expose the transaction hash in the `ExecutionFinalResult`?

workspaces/src/rpc/client.rs
shariffdev: I needed to access the `InMemorySigner::inner` method and also the `fetch_tx_nonce` which are private outside the workspace.
I found no clean way to construct the `SignedTransaction` on the examples side and added this convenience method.

frol: TL;DR: Let's remove `check_tx` and `signed_transaction`. `check_tx` is going to be removed from nearcore soon, and it does not make any sense in the context of workspaces-rs to have this method anyway.

It would be better to introduce a method to get a signed transaction after construction it via various methods - there are a number of helpers here that use `.transact().await` at the end, where we could introduce a new `.prepare_signed_transaction().await`, but it I don't see a demand for that, so we should not spend time on it right now.

workspaces/src/worker/impls.rs
frol: @ChaoticTempest kindly pinging you

frol (approved):

workspaces/src/worker/impls.rs
ChaoticTempest: hmm, why would this be an abstraction leak? `Worker` here is just the interface where we want to surface RPC APIs we deem safe to do so, or APIs that are not directly tied to an account/contract that requires signing a transaction with. `Worker` here also protects against unwanted RPC methods by constraining certain RPC APIs to certain networks like sandbox's `fast_forward`. Where do you imagine it would live otherwise?

examples/src/protocol_config.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    let block_reference = {
+        let hash = outcome.outcome().block_hash;
+        near_primitives::types::BlockReference::BlockId(near_primitives::types::BlockId::Hash(
+            near_primitives::hash::CryptoHash(hash.0),
+        ))
+    };
+
+    // NOTE: this API is under the "experimental" flag and no guarantees are given.
+    let protocol_config = worker.protocol_config(block_reference).await?;
```
ChaoticTempest, line 25: Note, this will make the supplying of `block_reference` inconsistent with: https://github.com/near/workspaces-rs/blob/35151975c165b693ec795841107978ab5e6800b6/examples/src/various_queries.rs#L90

It would be great if these methods were utilized with the following struct https://github.com/near/workspaces-rs/blob/35151975c165b693ec795841107978ab5e6800b6/workspaces/src/rpc/query.rs#L57-L87 so that we can keep things consistent and not have the user do all boilerplate each time

examples/src/receipt.rs
```diff
@@ -0,0 +1,40 @@
...
+
+        ReceiptReference {
+            receipt_id: near_primitives::hash::CryptoHash(
+                ids.pop().expect("expected at least one receipt id").0,
+            ),
+        }
+    };
+
+    // NOTE: this API is under the "experimental" flag and no guarantees are given.
+    let resp = worker.receipt(receipt_ref).await?;
```
ChaoticTempest, line 36: We don't need to expose the fact that we're using `ReceiptReference` since all it's taking is a `CryptoHash`. So, we can just have the signature `fn receipt(CryptoHash)` instead.

examples/src/changes.rs
```diff
@@ -0,0 +1,42 @@
...
+    };
+
+    let state_changes = {
+        StateChangesRequestView::ContractCodeChanges {
+            account_ids: vec![contract.id().clone()],
+        }
+    };
+
+    // NOTE: this API is under the "experimental" flag and no guarantees are given.
+    let res = worker.changes(block_ref, state_changes).await?;
```
ChaoticTempest, line 32: Same with this one and consistency with `Query` struct. We only need to supply `state_changes` while further methods can supply `block_ref` if it's needed with `Finality::Final` being the default when not supplied

shariffdev:
I'll put up a patch PR to address the reviews @ChaoticTempest

Issue #278, opened 2023-06-29
Improve error message when calling `json` on void function


agostbiro:
When calling `json()` on the execution result of function that doesn't return anything, the following error is generated:

```
Error: DataConversion

Caused by:
    EOF while parsing a value at line 1 column 0
```

We should catch this and create a better error message for the user that explains the problem.

Minimal code to reproduce assuming a contract with a `new` method that doesn't return anything:

```rust
let worker = workspaces::sandbox().await?;
let wasm = fs::read(/*WASM FILEPATH*/)?;
let contract = worker.dev_deploy(&wasm).await?;
contract.call("new").transact().await?.json()?;
```

PR #286, opened 2023-08-07
fix: improve error msg on calling `json` on void function

shariffdev:
Closes https://github.com/near/workspaces-rs/issues/278

The error returned on calling `json()` after execution of a void function is:
```
Error: DataConversion

Caused by:
    EOF while parsing a value at line 1 column 0
```

The change proposed gives more context to the error as such:
```
Error: DataConversion

Caused by:
    the function call returned an empty value, which cannot be parsed as JSON
```


shariffdev:
@frol This PR is ready to be reviewed

workspaces/src/result.rs
```diff
@@ -227,7 +227,20 @@ impl ExecutionFinalResult {
...
     pub fn json<T: serde::de::DeserializeOwned>(self) -> Result<T> {
-        self.into_result()?.json()
+        let val = self.into_result()?;
+        match val.json() {
+            Err(err) => {
+                // This catches the case: `EOF while parsing a value at line 1 column 0`
+                // for a function that doesn't return anything; this is a more descriptive error.
+                if *err.kind() == ErrorKind::DataConversion && val.value.repr.is_empty() {
+                    return Err(ErrorKind::DataConversion
+                        .custom("the Value from ExecutionOutcome is zero bytes"));
```
frol, line 237: I suggest a more clear error message:

```suggestion
                        .custom("the function call returned an empty value, which cannot be parsed as JSON"));
```

frol (changes requested):

frol (approved):

Issue #279, opened 2023-06-30
workspaces-rs should not require serde_json `arbitrary_precision` feature


frol:
https://github.com/near/workspaces-rs/blob/fd6e62bba2a060e31f5634322b511ebaf3428784/examples/Cargo.toml#L14-L15

This is an odd one, and given that [there are a ton of weird issues with `arbitrary_precision` feature](https://github.com/serde-rs/json/issues?q=is%3Aissue+is%3Aopen+arbitrary_precision), I would avoid using it.

PR #287, opened 2023-08-08
fix: rm-serde-arbitrary-precision-feat

shariffdev:
Closes https://github.com/near/workspaces-rs/issues/279

The `arbitrary_precision` feature is removed from serde_json in the Cargo.toml


itsyaasir:

itsyaasir:
@frol  This PR is ready to be reviewed

shariffdev:
@frol This is ready to be reviewed.

frol:
@shariffdev Why was it used in the first place? Does it break anything when this feature is dropped? cc @ChaoticTempest

shariffdev:
> @shariffdev Why was it used in the first place? Does it break anything when this feature is dropped? cc @ChaoticTempest

I did not find any example that broke without the feature. Please advise @ChaoticTempest

frol (changes requested):
`arbitrary_precision` was introduced in #35. Please, run ref-finance example with this PR applied and confirm that it still works just fine, and let's merge it.

shariffdev:
I have pushed a small fix, everything is looking good now!

examples/src/ref_finance.rs
```diff
@@ -71,7 +71,7 @@ async fn create_wnear(owner: &Account, worker: &Worker<Sandbox>) -> anyhow::Resu
         .call(wnear.id(), "new")
         .args_json(json!({
             "owner_id": owner.id(),
-            "total_supply": parse_near!("1,000,000,000 N"),
+            "total_supply": parse_near!("1,000,000,000 N").to_string(),
         }))
```
frol, line 75: `wrap.near` contract `new` method does not expect any arguments O_o

https://github.com/near/core-contracts/blob/3f3170fce91ff4d8c6ee9d15683f2d4dfe1275cf/w-near/src/lib.rs#L36

Let's remove `args_json` completely.

shariffdev: OK, let me try!

shariffdev: It works okay!

frol (approved):

PR #288, opened 2023-08-10
feat: ExecutionOutcome.status should be public

robert-zaremba:
Today, the only way to check the transaction error content is to serialize it with JSON or format it with default display formatter. 
Moreover it's hard to check an error in a particular receipt.
Example: https://github.com/near-ndc/voting-v1/blob/8f1c14a/elections/tests/iah.rs#L169

Potential workaround would be to make `Error.repr` public. But the flow will be even more complicated:
```
    let failures = res.receipt_failures();
    for f in failures {
        let err = f.into_result().err().unwrap();
        match err.repr { // <-- Doesn't work because repr is private
            // moreover the following is too complicated: requires the knowledge of ErrorRepr
            ErrorRepr::Custom(kind, error) => { /* match and check erro, which is a Box */ }
            _ ....
        };
    }
```

### DIFF
--- a/workspaces/src/result.rs
+++ b/workspaces/src/result.rs
@@ -408,7 +408,7 @@ pub struct ExecutionOutcome {
     /// for receipt this is receiver_id.
     pub executor_id: AccountId,
     /// Execution status. Contains the result in case of successful execution.
-    pub(crate) status: ExecutionStatusView,
+    pub status: ExecutionStatusView,
 }
 
 impl ExecutionOutcome {


ChaoticTempest:
We didn't expose `status` since the errors coming from nearcore are very likely to change until they are stabilized with `near-primitives` version `1.0`. This increases the likelihood of breaking changes when it comes time to upgrade dependencies. It also increases the churn when it comes to having someone update their workspaces version. Alternatively, if you want to match against the errors, you can call `downcast` directly on the errors to get the specific near_primitives error you want

frol:
@robert-zaremba Have you tried `downcast_ref` as @ChaoticTempest suggested?

```rust
if let Some(err) = f.into_result().unwrap_err().source().unwrap().downcast_ref::<near_primitives::errors::TxExecutionError>() {
    dbg!(err);
}
```

frol:
@robert-zaremba Did my snippet help? Should we close this for now (until we are ready to create an API-stable version of the outcome status)?

josemariasosa:
@frol 

I tried using your snippet, but I'm getting this error:

```
called `Option::unwrap()` on a `None` value
```

I added these two lines of code:

```rs
use std::error::Error;
use near_primitives;
```

And these are my dev-dependencies:

```
[dev-dependencies]
anyhow = "1.0"
near-sdk = { version = "5.5", features = ["unit-testing"] }
near-workspaces = { version = "0.14.0", features = ["unstable"] }
tokio = { version = "1.12.0", features = ["full"] }
serde_json = "1"
near-gas = "0.3.0"
near-primitives = "0.28.0"
```

After testing some smart contracts, I noticed that checking the result of a call using the `is_failure()` method is not particularly useful unless I explicitly compare the returned error with the specific error I expect. This contrasts with the `is_success()` method, which directly tells me if a call succeeded.

In Hardhat, there is a convenient function `revertedWithCustomError` that allows you to check if a transaction failed with a specific custom error. Is there any equivalent beside parsing the low level error?

Thanks!

josemariasosa:
UPDATE

Not the most elegant solution, but it works. If anybody arrives here with the same issue I had, this piece of code might work for you.

```rs
let result = alice.call(protocol.borrow_verde_contract.id(), "withdraw_collateral")
    .args_json(json!({
        "collateral_id": protocol.st_near_contract.id(),
        "collat_amount": U128::from(0),
    }))
    .deposit(NearToken::from_yoctonear(1))
    .max_gas()
    .transact()
    .await?;
assert!(result.is_failure() && result.into_result().unwrap_err().to_string().contains("InvalidZeroWithdraw"));
```

PR #290, opened 2023-08-21
chore/transfer-near-test

shariffdev:
Addresses TODO: https://github.com/near/workspaces-rs/blob/cda1dc4e481e91bb17d95053b42d3b584cf2a95b/workspaces/src/rpc/client.rs#L185

The PR adds a test case for the `transfer-near` rpc method.


shariffdev:

workspaces/tests/account.rs
```diff
@@ -33,3 +33,24 @@ async fn test_subaccount_creation() -> anyhow::Result<()> {
...
+
+#[test(tokio::test)]
+async fn test_transfer_near() -> anyhow::Result<()> {
+    let worker = workspaces::sandbox().await?;
+    let (alice, bob) = (
+        worker.dev_create_account().await?,
+        worker.dev_create_account().await?,
+    );
+
+    // transfer 500_000_000 token from alice to bob
```
frol, line 45: Please, add asserts on that alice's balance has changed (one assert before, and one after the transfer)

frol (changes requested):

frol (approved):

Issue #272, opened 2023-05-05
A lot of noisy message: `Updated the logging layer according to `log_config.json`


mrLSD:
It is a lot of noisy log messages in the output:
```
Updated the logging layer according to `log_config.json`
```

Link to GitHub CI:
https://github.com/aurora-is-near/aurora-workspace/actions/runs/4893433957/jobs/8736389991#step:5:473

`no-capture` flag is not enabled.

Tested in 3 completely different projects. Same log message is in the output.

itegulov:
Seems like this is happening because of https://github.com/near/nearcore/pull/8810. Not sure if there is a good way to silence this other than just not inheriting stdout from sandbox. @ChaoticTempest thoughts?

ChaoticTempest:
yeah looks like we're just inheriting stdout by default for sandbox processes. A couple options we have for silencing it:
- set the sandbox proccess stdout to `Stdio::null` which just directs it to `/dev/null`, but nearcore might have informative stdout/stderr related info if it crashes or something can be added in the future, so probably not the best option.
- Since we have sandbox/network-builders now, we can have an extra parameter there saying whether or not to include stdout, with the default being not to include

itsyaasir:
👋🏾  As per our discussion with @frol , I would like to take up this issue.

frol:
Another option I would consider - fix it on nearcore side - avoid using println (use logging instead)

itsyaasir:
> Another option I would consider - fix it on nearcore side - avoid using println (use logging instead)

Yeah, I think I will look into this.

itsyaasir:
I have confirmed that the messages are no longer printed in the `stdout`. Attached is a screenshot with two commands.
- The first command is running with custom sandbox built against this branch, as you can see there's no println messages
- The second command is using the default sandbox and the print messages are there

PS:- The default version uses Commit hash for the sandbox "versioning", we will have to update the hash if we want to use the new sandbox which doesn't have the logging messages.

![Screenshot 2023-07-07 at 9 08 57 AM](https://github.com/near/nearcore/assets/7762687/5bdc3b99-2d92-4175-9c2f-537bb4aa2a44)

mrLSD:
@itsyaasir is there a way to test it with `workspaces-rs`? Maybe git hash?

P.F. Out of discussion: what's the cool plugin for `zsh`? ))

itsyaasir:
[...]

PR #293, opened 2023-08-30
chore: upgrade to stable toolchain

shariffdev:
The issue involving wasm contract builds for Rust toolchain 1.70 and later: https://github.com/near/nearcore/issues/9143 has been resolved 
Should be ok to update the toolchain.

<s>Are we updating to toolchain `1.70` or ok to use `stable`?</s>

### DIFF
--- a/workspaces/build.rs
+++ b/workspaces/build.rs
@@ -5,7 +5,7 @@ fn main() {
     let env_bin = std::env::var("NEAR_SANDBOX_BIN_PATH").is_ok();
     if !doc_build && !env_bin && cfg!(feature = "install") {
         // TODO Update commit to stable version once binaries are published correctly
-        // Commit: https://github.com/near/nearcore/commit/dce4c514f60a7f7981f00290268af4faf7eec55c
+        // Commit: https://github.com/near/nearcore/commit/eb2bbe1c3f51912c04462ce988aa496fab03d60e
         near_sandbox_utils::install_with_version("master/eb2bbe1c3f51912c04462ce988aa496fab03d60e")
             .expect("Could not install sandbox");
     }
--- a/workspaces/build.rs
+++ b/workspaces/build.rs
@@ -5,6 +5,8 @@ fn main() {
     let env_bin = std::env::var("NEAR_SANDBOX_BIN_PATH").is_ok();
     if !doc_build && !env_bin && cfg!(feature = "install") {
         // TODO Update commit to stable version once binaries are published correctly
-        near_sandbox_utils::install().expect("Could not install sandbox");
+        // Commit: https://github.com/near/nearcore/commit/dce4c514f60a7f7981f00290268af4faf7eec55c
+        near_sandbox_utils::install_with_version("master/eb2bbe1c3f51912c04462ce988aa496fab03d60e")
+            .expect("Could not install sandbox");
     }
 }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,9 +31,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        # TODO: Bring `stable` back to the matrix once Rust 1.70.0+ compatibility is resolved in nearcore:
-        # https://github.com/near/nearcore/issues/9143
-        toolchain: [1.69.0]
+        toolchain: [stable]
 
     runs-on: ${{ matrix.platform }}
     steps:
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 ## Requirements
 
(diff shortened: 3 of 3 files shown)


shariffdev:
@frol This is for updating the toolchain to stable. The wasm contract should now be buildable with the latest Rust Stable Version

shariffdev:
Quick question, I have updated the build.rs file to use sandbox with a newer commit from the nearcore repo, and I changed `install` instead I used `install_with_version`. Is this ok ? or shall I update the sandbox utils crate ?

workspaces/build.rs
```diff
@@ -5,6 +5,8 @@ fn main() {
     let env_bin = std::env::var("NEAR_SANDBOX_BIN_PATH").is_ok();
     if !doc_build && !env_bin && cfg!(feature = "install") {
         // TODO Update commit to stable version once binaries are published correctly
-        near_sandbox_utils::install().expect("Could not install sandbox");
+        // Commit: https://github.com/near/nearcore/commit/dce4c514f60a7f7981f00290268af4faf7eec55c
+        near_sandbox_utils::install_with_version("master/eb2bbe1c3f51912c04462ce988aa496fab03d60e")
+            .expect("Could not install sandbox");
```
frol, line 10: Please, update near_sandbox_utils to use the most recent stable release of nearcore.

frol (changes requested):
Other than the comment above, we are good to go. Given that it will require near_sandbox_utils publishing, I think we can use `install_with_version` here for the time being, but let's pick the most recent stable release in nearcore instead of some random recent commit.

shariffdev:
@frol if we use the most recent stable release which is 1.35, it doesn't include the fixes for this https://github.com/near/workspaces-rs/issues/272, It is either I use this recent commit or edit the sandbox and use a more recent [hash commit](https://github.com/near/sandbox/blob/ee06569991da4a3a69f3cc575715a3fdc06935da/crate/src/lib.rs#L15)

Or we can forgo this issue #272 entirely, until it is included in the next release.

frol:
I see, indeed, I hoped there was another release since that time. I think we should be fine using more recent nearcore snapshot for now as workspaces-rs only needs to run it in a local node mode, so there is no risk of incompatibility, insecurity, or instability of it. That said, probably sandbox-utils does not need to be updated to some random hash commit, let's upgrade it to the most recent release of nearcore while workspaces-rs would use some custom commit hash.

shariffdev:
Working on the sandbox utils now.

shariffdev:
> @shariffdev @frol Currently when we use workspace-rs that issue [near/nearcore/issues/9143](https://github.com/near/nearcore/issues/9143) is still reproducible. For example for this [aurora-is-near/aurora-engine/pull/832](https://github.com/aurora-is-near/aurora-engine/pull/832).
> 
> Does this PR solve issue also for nightly (current) versions?

Yes this should solve the issue you mentioned.

workspaces/build.rs
shariffdev: fixed

frol (approved):

shariffdev:
> @shariffdev this PR was merged. But when I link the latest main hash 8d93197  as a dependency, near/nearcore/issues/9143 is still reproducible for stable and nightly. Any ideas?
I've looked into this with an example contract in a fresh project dir using the stable toolchain and things are fine. Could you have missed something?

mrLSD:
> Could you have missed something?

@shariffdev 
1. For the building contract I used `near-sdk = "4.1.1"`
2. For tests: `workspaces = { git = "https://github.com/near/workspaces-rs.git", rev = "8d93197" }`
3. `rustc 1.72.0 (5680fa18f 2023-08-23)`

All tests failed with `Deserialization` error.

shariffdev:
@mrLSD Try doing a `cargo clean` I recreated your issue and that did it for me

mrLSD:
@shariffdev 
`cargo clean && rm -rf target/` - doesn't work.

```
status: Failure(ActionError(ActionError { index: Some(0), kind: FunctionCallError(CompilationError(PrepareError(Deserialization))) })),
```

shariffdev:
I'm out of ideas but still looking into it. Is there a chance the code is open and I can take a look at it?

mrLSD:
@shariffdev 
It fails [here](https://github.com/aurora-is-near/release-repository/blob/main/tests/utils.rs#L53)

How it was tested:
1. Toolchain set to stable (or nightly-2023-08-24)
2. Set Cargo.toml: `workspaces = { git = "https://github.com/near/workspaces-rs.git", rev = "8d93197" }`
3. `cargo update`
4. Run build and tests: `make test`

I tested for our projects:

https://github.com/aurora-is-near/aurora-eth-connector/tree/master/eth-connector-tests

https://github.com/aurora-is-near/aurora-engine/tree/develop/engine-tests

Same issues.

Issue #229, opened 2022-11-02
`fast_forward` misleading name


austinabell:
Intuitively I would expect `fast_forward` to progress the state transitions as if the time had passed naturally. Opening issue because I (and potentially others) might be surprised to find out it just progresses the block height, timestamp, and epoch and leaves the current in-flight transactions as-is.

My argument for why it should progress the transactions is that testing transaction ordering is very non-deterministic and relies on manually waiting in tests. Having `fast_forward` progress the transactions doesn't solve the non-determinism, but can make it much harder to hit.

At very least, the docs for `fast_forward` should indicate that any receipts that are scheduled and have not been executed will not be progressed during this call, I think.

PR #299, opened 2023-09-11
chore: improve fast_forward docs

shariffdev:
Closes https://github.com/near/workspaces-rs/issues/229

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,7 +289,8 @@ Following that we will have to init the contract again with our own metadata. Th
 
 ### Time Traveling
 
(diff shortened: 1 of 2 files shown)


frol (approved):
> At very least, the docs for fast_forward should indicate that any receipts that are scheduled and have not been executed will not be progressed during this call, I think.

I think this requirement is fulfilled, but we may get back to it in the context of more end-to-end DevX to make sure that the function name is not misleading (we cannot expect developers to read all the documentation)

Issue #196, opened 2022-09-20
`keystore_path` in `Info` doesn't appear to be in use


joshuajbouw:
It would be quite handy to be able to import your keys and have access to them in the `Sandbox`, `Testnet` and `Mainnet` workers. However, it only sets the path for these credentials but does not load the keys themselves.

## Solution

I do not exactly see an actual benefit to holding this information as it isn't used anywhere in the code. What would be nice if there was a key loader where you can have access to your accounts however!

frol:
@joshuajbouw What you are asking is a bad practice of storing keys in plain text on disk, and if you implement such a test with the workspaces-rs, you will need to replicate this insecure design on CI as well. I am hesitant to promote such bad practices to the community.

@shariffdev Since you jumped on it, I recommend you think about the problem space more holistically, and design a solution that would be convenient, but not compromised on security.

PR #302, opened 2023-09-13
chore: cleanup unused `keystore_path`

shariffdev:
The `keystore_path` is unused, there's an issue [here](https://github.com/near/workspaces-rs/issues/196) indicating the need for a key loader. 

The keyloader is one of those features that might be rarely used but while reviewing you can nudge me to work on it, I had already started a wip implementation.


itsyaasir:
@frol This can be reviewed

frol:
Let's make a step back and think how it should be actually implemented: https://github.com/near/near-workspaces-rs/issues/196#issuecomment-1747506376

shariffdev:
@frol is this PR headed in the right direction? I reasoned testnet and localnet accounts are ok as plain text maybe weigh in on that?

PR #306, opened 2023-09-18
feat: api key and custom network object

shariffdev:
Inherits work from https://github.com/near/near-workspaces-rs/pull/190. Resolving https://github.com/near/near-workspaces-rs/issues/184


workspaces/src/rpc/client.rs
```diff
@@ -61,15 +61,20 @@ pub struct Client {
 }
 
 impl Client {
-    pub(crate) fn new(rpc_addr: &str) -> Self {
+    pub(crate) fn new(rpc_addr: &str) -> Result<Self> {
         let connector = JsonRpcClient::new_client();
-        let rpc_client = connector.connect(rpc_addr);
+        let mut rpc_client = connector.connect(rpc_addr);
+        if let Ok(api_key) = std::env::var("NEAR_RPC_API_KEY") {
```
frol, line 67: I find it quite a bit ugly to configure anything in library via env variables. What if I need to connect to two different APIs (e.g. testnet and mainnet) that have different API keys? I feel that `Client` type should just have a separate method to configure API key for `self.rpc_client`, instead of trying to configure it in the `new` method while trying to sneak it in via env variable.

README.md
```diff
@@ -344,5 +344,7 @@ These environment variables will be useful if there was ever a snag hit:
 
 - `NEAR_RPC_TIMEOUT_SECS`: The default is 10 seconds, but this is the amount of time before timing out waiting for a RPC service when talking to the sandbox or any other network such as testnet.
 - `NEAR_SANDBOX_BIN_PATH`: Set this to our own prebuilt `neard-sandbox` bin path if we want to use a non-default version of the sandbox or configure nearcore with our own custom features that we want to test in workspaces.
+<<<<<<< HEAD
```
frol, line 347: nit: there is a merge artifact.

frol (changes requested):

shariffdev:
@frol Ready to be reviewed

frol (approved):

PR #309, opened 2023-09-21
fix!: add testnet create-tla exec details

shariffdev:
Cleaning up a chore from https://github.com/near/near-workspaces-rs/pull/168/files#r920173526

- [x] Checking if should be part of `near-jsonrpc-client`: https://github.com/near/near-workspaces-rs/blob/d063b4f2e03308d87fe12cacae1eefad7763edea/workspaces/src/rpc/tool.rs#L45
The `https://helper.testnet.near.org` network has a one-off use case for the account creation. This can be left as is without adding it as an rpc which are all network generic.


shariffdev:
@frol ready for review

shariffdev:
TLA creation not possible going forward.

PR #310, opened 2023-09-25
feat: keyloader interopable with near-cli

shariffdev:
Closes: https://github.com/near/near-workspaces-rs/issues/196

It helps load keys saved using the `near-cli` cross-platform keychain,  can also just used to set and get account credentials from keychain. 

Using `unstable` feature flag, it offers interoperability with `near-cli` which is not yet v1.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,5 +47,10 @@ jobs:
       run: rustup target add wasm32-unknown-unknown
     - name: Check with stable features
       run: cargo check --verbose
+    - name: Install required dependencies (linux)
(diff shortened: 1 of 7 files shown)


frol:
@shariffdev Using system keychain is not going to work on CI, and also not very productive locally, and the original issue #196 probably suggested to use a folder with plain-text keys stored on disk, which is how near-cli JS does it. I am hesitant to support that as it promotes bad practices. Let me elaborate more on the issue itself.

frol:
P.S. the "keystore" referred in near-cli is called "legacy keystore" in near-cli-rs

PR #313, opened 2023-09-28
feat: mitigate infinite running tests

shariffdev:
This PR mitigates the issue of infinite running tests for the [Linux CI issue](https://github.com/near/near-workspaces-rs/issues/312) by using retries and timeouts to terminate tests.


frol:
Well, it did not fix it, and also we cannot really recommend users to use such unstable testing framework, so we really need to find a better solution.

shariffdev:
Thanks for the feedback, we'll explore alternative solutions. 🙏

PR #318, opened 2023-10-04
refactor!: renamed crate to near-workspaces to avoid confusion with Cargo workspaces; imports should now use `near_workspaces` instead of just `workspaces`

frol:


ChaoticTempest:
@frol just wanted to let you know that we have the `near` package currently squatted on crates.io. It would be great if that package could directly link to all the other libraries too, just like a top level lib. Then we'd be able to do the following pretty nicely:
```
use near::workspaces; 
use near::sdk;
use near::json_rpc_client;
```

frol:
@ChaoticTempest since rarely anyone want workspaces, sdk, and json_rpc_client in a single package, I don't think I will go that route, but I do remember about that package and I was planning to use it somehow...

PR #323, opened 2023-10-06
fix: satisfy conditions for semver-check

shariffdev:


frol:
@shariffdev I am not familiar with UnwindSafe, so I have several questions in my head:

1. Why did we lose auto-trait implementation there in the first place?
2. Is it safe to just impl explicitly?
3. Do we need to impl it explicitly?

Could you dig into it?

shariffdev:
Sure, looking into it.

shariffdev:
> Why did we lose auto-trait implementation there in the first place?
> Do we need to impl it explicitly?

The GasHook introduced did not implement the auto-traits because it does not fall under the rules the compiler uses for auto implementation. We had to add `Sync` and `Send` explicitly which were also not auto-implemented. Reference: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits

> Is it safe to just impl explicitly?

The traits behave similarly in implementation to `Send` and `Sync` which we have also explicitly implemented. Reference: https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html

frol (approved):
Thanks for diving into it! This looks reasonable now

frol:
Oh, it should have targeted `main` branch, and then release-plz would pick it up from there.

shariffdev:
I missed that! Is anything to be done or we leave it?

PR #328, opened 2023-10-17
chore: apply suggestions proposed by clippy

aleksuss:
The motivation to propose these changes is to add the possibility of using this crate in projects which are checked with clippy with lints `all` and `nursery`.

### DIFF
--- a/workspaces/src/result.rs
+++ b/workspaces/src/result.rs
@@ -399,7 +399,7 @@ impl ViewResultDetails {
 
 impl From<CallResult> for ViewResultDetails {
     fn from(result: CallResult) -> Self {
-        ViewResultDetails {
+        Self {
             result: result.result,
             logs: result.logs,
         }
@@ -525,7 +525,7 @@ impl Value {
 
 impl From<ExecutionOutcomeWithIdView> for ExecutionOutcome {
     fn from(view: ExecutionOutcomeWithIdView) -> Self {
-        ExecutionOutcome {
+        Self {
             transaction_hash: CryptoHash(view.id.0),
             block_hash: CryptoHash(view.block_hash.0),
             logs: view.outcome.logs,
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -482,7 +482,6 @@ pub(crate) async fn access_key(
 /// Fetches the transaction nonce and block hash associated to the access key. Internally
 /// caches the nonce as to not need to query for it every time, and ending up having to run
 /// into contention with others.
-#[allow(clippy::significant_drop_in_scrutinee)]
 async fn fetch_tx_nonce(
     client: &Client,
     cache_key: &(AccountId, near_crypto::PublicKey),
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -81,10 +81,7 @@ impl Client {
     pub(crate) async fn query_broadcast_tx(
         &self,
         method: &methods::broadcast_tx_commit::RpcBroadcastTxCommitRequest,
-    ) -> MethodCallResult<
-        FinalExecutionOutcomeView,
-        near_jsonrpc_primitives::types::transactions::RpcTransactionError,
-    > {
+    ) -> MethodCallResult<FinalExecutionOutcomeView, RpcTransactionError> {
         retry(|| async {
             let result = self.rpc_client.call(method).await;
             match &result {
@@ -124,16 +121,16 @@ impl Client {
 
     pub(crate) async fn query_nolog<M>(&self, method: M) -> MethodCallResult<M::Response, M::Error>
     where
-        M: methods::RpcMethod,
+        M: methods::RpcMethod + Send + Sync,
     {
         retry(|| async { self.rpc_client.call(&method).await }).await
     }
 
     pub(crate) async fn query<M>(&self, method: M) -> MethodCallResult<M::Response, M::Error>
     where
-        M: methods::RpcMethod + Debug,
-        M::Response: Debug,
-        M::Error: Debug,
+        M: methods::RpcMethod + Debug + Send + Sync,
+        M::Response: Debug + Send,
+        M::Error: Debug + Send,
     {
         retry(|| async {
             let result = self.rpc_client.call(&method).await;
@@ -456,7 +453,7 @@ impl Client {
 
 pub(crate) async fn access_key(
     client: &Client,
-    account_id: near_primitives::account::id::AccountId,
+    account_id: AccountId,
     public_key: near_crypto::PublicKey,
 ) -> Result<(AccessKeyView, CryptoHash)> {
     let query_resp = client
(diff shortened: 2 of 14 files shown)


workspaces/src/network/sandbox.rs
```diff
@@ -19,7 +19,7 @@ use crate::types::{AccountId, Balance, InMemorySigner, SecretKey};
 use crate::{Account, Contract, Network, Worker};
 
 // Constant taken from nearcore crate to avoid dependency
-pub(crate) const NEAR_BASE: Balance = 1_000_000_000_000_000_000_000_000;
+pub const NEAR_BASE: Balance = 1_000_000_000_000_000_000_000_000;
```
frol, line 22: This is ugly const that I would prefer not to expose as part of the public interface. https://github.com/near/near-token will hopefully emerge next week and that would be a better way to deal with NEAR tokens as a type

workspaces/src/network/variants.rs
```diff
@@ -5,7 +5,7 @@ use crate::types::{AccountId, KeyType, SecretKey};
 use crate::{Account, Contract, Worker};
 use async_trait::async_trait;
 
-pub(crate) const DEV_ACCOUNT_SEED: &str = "testificate";
+pub const DEV_ACCOUNT_SEED: &str = "testificate";
```
frol, line 8: What is the motivation to make it public?

workspaces/src/result.rs
```diff
@@ -334,7 +334,7 @@ impl ExecutionSuccess {
 
 impl<T> ExecutionResult<T> {
     /// Returns just the transaction outcome.
-    pub fn outcome(&self) -> &ExecutionOutcome {
+    pub const fn outcome(&self) -> &ExecutionOutcome {
```
frol, line 337: Since you cannot get outcome without making a request, I believe it is quite pointless to make this method `const`. Removing `const` in the future will be breaking change, so I would avoid unnecessary `const` markers.

P.S. Apply the same reasoning to all the other methods as I feel that half of the `const` methods should not be `const` really.

workspaces/src/rpc/client.rs
```diff
@@ -49,8 +49,8 @@ use crate::result::Result;
 use crate::types::{AccountId, InMemorySigner, Nonce, PublicKey};
 use crate::{Network, Worker};
 
-pub(crate) const DEFAULT_CALL_FN_GAS: NearGas = NearGas::from_tgas(10);
-pub(crate) const DEFAULT_CALL_DEPOSIT: Balance = 0;
+pub const DEFAULT_CALL_FN_GAS: NearGas = NearGas::from_tgas(10);
+pub const DEFAULT_CALL_DEPOSIT: Balance = 0;
```
frol, line 53: What is the motivation for exposing these internal consts?

workspaces/src/rpc/client.rs
```diff
@@ -124,16 +121,16 @@ impl Client {
 
     pub(crate) async fn query_nolog<M>(&self, method: M) -> MethodCallResult<M::Response, M::Error>
     where
-        M: methods::RpcMethod,
+        M: methods::RpcMethod + Send + Sync,
```
frol, line 124: What is the motivation for adding extra constraints?

workspaces/src/rpc/client.rs
```diff
@@ -527,7 +525,7 @@ where
     Retry::spawn(retry_strategy, task).await
 }
 
-pub(crate) async fn send_tx(
+pub async fn send_tx(
```
frol, line 528: Making these methods public will require more time to review and make sure that the interfaces are aligned and documented.

workspaces/src/worker/impls.rs
```diff
@@ -144,8 +144,27 @@ where
...
+    }
+
+    pub fn gas_price(&self) -> Query<'_, GasPrice> {
+        Query::new(self.client(), GasPrice)
+    }
+}
+
+impl<T> Worker<T>
+where
+    T: NetworkClient + Send + Sync + ?Sized,
```
frol, line 164: What is the motivation here? Why some methods require more constraints than others?

frol (changes requested):
Thanks for bringing this up, but we need to understand motivation here. There are some obviously good changes, but there are also a lot of controversial ones, especially around promoting `pub(crate)` to `pub`. Please, provide more details in the comments

workspaces/src/worker/impls.rs
aleksuss: Because in this block, the methods are asynchronous and with such bounds, we tell the compiler that type `T` will be moved between threads safely.

aleksuss:
As described in the title, the changes have been proposed by the clippy linter. The point why clippy suggests making some methods and constants public is because they are defined in private modules, and these changes don't affect public API.

workspaces/src/network/sandbox.rs
aleksuss: It won't be exposed because the `sandbox` module is private.

workspaces/src/network/config.rs
```diff
@@ -57,7 +57,7 @@ where
 }
 
 /// Set extra configs for the sandbox defined by workspaces.
-pub(crate) fn set_sandbox_configs(home_dir: impl AsRef<Path>) -> Result<()> {
+pub fn set_sandbox_configs(home_dir: impl AsRef<Path>) -> Result<()> {
```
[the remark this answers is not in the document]
aleksuss: It's a suggestion by clippy. The motivation is that the `pub(crate)` is unnecessary because the `config` module is private.

workspaces/src/rpc/client.rs
aleksuss: Because `M` could be shared between threads.

workspaces/src/network/sandbox.rs
frol: This is a false positive: https://github.com/rust-lang/rust-clippy/issues/8319

frol:
@aleksuss The compilation error due to the Send + Sync is real and indeed has to be addressed (please, also add a test).

I have disabled `clippy::missing_const_for_fn` and `clippy::redundant_pub_crate` lints as they produce false positives:

https://github.com/near/near-workspaces-rs/blob/b6f16a74e7ecb9a9f2b79a1a6a9d8533565427c0/workspaces/src/lib.rs#L7

Could you, please, revert the changes that added `const fn` and change visibility from `pub(crate)` to `pub`, and resolve merge conflicts?

workspaces/src/rpc/client.rs
```diff
@@ -485,6 +482,7 @@ pub(crate) async fn access_key(
 /// Fetches the transaction nonce and block hash associated to the access key. Internally
 /// caches the nonce as to not need to query for it every time, and ending up having to run
 /// into contention with others.
+#[allow(clippy::significant_drop_in_scrutinee)]
```
frol, line 485: This should not be needed after my fix #334

```suggestion
```

frol (approved):

frol:
@aleksuss Does your code work if you compile it using your forked near-workspaces?

aleksuss:
@frol yes, now it compiles without errors.

frol:
@aleksuss near-workspaces [0.9 release](https://github.com/near/near-workspaces-rs/releases/tag/near-workspaces-v0.9.0) is published! Thank you for contributing!

PR #329, opened 2023-10-17
chore: cleanup internals

shariffdev:
~Also resolving https://github.com/near/near-workspaces-rs/issues/153~

### DIFF
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -9,7 +9,6 @@ use tokio::sync::RwLock;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
-use near_crypto::Signer;
 use near_jsonrpc_client::errors::{JsonRpcError, JsonRpcServerError};
 use near_jsonrpc_client::methods::health::RpcStatusError;
 use near_jsonrpc_client::methods::tx::RpcTransactionError;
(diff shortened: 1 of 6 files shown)


workspaces/src/rpc/client.rs
```diff
@@ -559,22 +558,13 @@ pub(crate) async fn send_batch_tx_and_retry(
...
             &cache_key,
-            SignedTransaction::from_actions(
-                nonce,
-                signer.account_id.clone(),
-                receiver_id.clone(),
-                &signer as &dyn Signer,
-                actions.clone(),
-                block_hash,
-            ),
+            signed_transaction(signer, actions.clone(), nonce, block_hash, receiver_id),
```
frol, line 567: I prefer constructor method (`from_actions` or maybe other name) on SignedTransaction over a standalone function as it is harder to discover the function. What is exactly the motivation behind this change?

frol (changes requested):

workspaces/src/rpc/client.rs
shariffdev: The function is a reimplementation of `from_actions` without requiring `Signer trait` as an argument. The hint I took [here](https://github.com/near/near-workspaces-rs/issues/153) is that we can condense that implementation and remove the reliance on the trait.
I could name it better.

workspaces/src/types/mod.rs
```diff
@@ -528,3 +521,26 @@ impl From<Finality> for near_primitives::types::BlockReference {
...
+        nonce,
+        receiver_id: receiver_id.clone(),
+        block_hash: near_primitives::hash::CryptoHash(block_hash.0),
+        actions,
+    };
+
+    SignedTransaction::new(
+        signer.secret_key.0.sign(tx.get_hash_and_size().0.as_ref()),
+        tx,
+    )
```
ChaoticTempest, line 545: you shouldn't need to call into `SignedTransaction::new`. A `tx.sign(&signer)` is simpler

frol (approved):

PR #337, opened 2023-11-07
chore: switch to using subaccounts & deprecate tla creation

shariffdev:


shariffdev:
@frol Ready for review

workspaces/tests/gas_meter.rs
```diff
@@ -315,3 +228,61 @@ async fn test_dropped_gas_meter() -> anyhow::Result<()> {
...
+                }),
+        )
+        .unwrap();
+
+    Ok(account.into_result()?)
+}
+
+const fn as_near(gas: Gas) -> NearToken {
+    NearToken::from_yoctonear(gas.as_gas() as u128)
+}
```
frol, line 288: What is going on here? You should not turn Gas into NearToken. Those are completely different things.

workspaces/tests/gas_meter.rs
```diff
@@ -286,21 +197,23 @@ async fn test_gas_meter_batch_tx() -> anyhow::Result<()> {
...
-
-    // analogous to: worker.dev_create_account().await?;
-    let account = {
-        let (id, sk) = worker.dev_generate().await;
-        let account = worker.create_tla(id.clone(), sk).await?;
-        total_gas += account.details.total_gas_burnt.as_gas();
-
-        account.into_result()?
-    };
+    let mut total_gas = NearToken::from_yoctonear(0);
```
frol, line 200: Gas has to be Gas, not NearToken

workspaces/src/types/mod.rs
```diff
@@ -44,6 +44,9 @@ pub type BlockHeight = u64;
 /// Shard index, from 0 to NUM_SHARDS - 1.
 pub type ShardId = u64;
 
+// Constant taken from nearcore crate to avoid dependency
+pub const DEFAULT_DEPOSIT: NearToken = NearToken::from_near(100);
```
frol, line 48: There is no sense in such a public constant. There is no such a thing like "default deposit" in NEAR protocol.

workspaces/src/network/variants.rs
```diff
@@ -15,6 +15,7 @@ pub trait NetworkInfo {
     fn info(&self) -> &Info;
 }
 
+#[deprecated = "only the registrar can create top level accounts as of Protocol >=64"]
```
frol, line 18: Top-level accounts creation is not deprecated, so I am not completely convinced that we should mark this trait as deprecated

frol (changes requested):

workspaces/src/types/mod.rs
shariffdev: I moved this over to types crate: https://github.com/near/near-workspaces-rs/blob/51e0ead85556c57f3bd41398cd75025ad51c33b3/workspaces/src/network/sandbox.rs#L22

workspaces/src/network/variants.rs
shariffdev: I added that as an attempt to discourage usage, maybe we use `warn`?

workspaces/src/types/mod.rs
shariffdev: > There is no sense in such a public constant. There is no such a thing like "default deposit" in NEAR protocol.

Updated the docs and renamed the constant to reflect that.

frol:
Resolved with #360

PR #339, opened 2023-11-21
Remove interop_sdk from defaults

iho:


workspaces/src/types/mod.rs
```diff
@@ -144,7 +144,7 @@ impl PublicKey {
             },
         )?))
     }
-
+    #[allow(dead_code)]
```
frol, line 147: Let’s be more explicit and hide it behind the interop_sdk feature flag instead of allowing dead code

frol (changes requested):
Thanks. This should fix the issues with using two incompatible versions of near-sdk-rs.

frol (approved):

PR #343, opened 2024-01-03
fix: patch near-sdk for broken build

shariffdev:
Getting `error: no matching package named parity-secp256k1 found` with main. 
We patch the error in this PR, meanwhile.


frol:
Cargo does not allow to publishing crates with git dependencies, so this PR is not mergeable. I don't think there is a value in keeping this PR open, so I am closing it.

PR #347, opened 2024-01-25
refactor: Updated cargo-near to "0.5.2"

FroVolod:


workspaces/Cargo.toml
```diff
@@ -14,7 +14,7 @@ async-trait = "0.1"
 base64 = "0.21"
 bs58 = "0.5"
 cargo_metadata = { version = "0.18", optional = true }
-cargo-near = "0.3.1"
+cargo-near = "0.5.1"
```
frol, line 17: Since we don't need ledger support, let's disable it and then revert the changes in the test.yml.

```suggestion
cargo-near = { version = "0.5.1", default-features = false }
```

frol (approved):

Issue #156, opened 2022-06-23
How to increase the TLA balance in Sandbox?


Eduardogbg:
I've read the docs and the source and it seems to me that there's no way to change how much initial balance the Sandbox TLA gets when created. That makes it impossible for me to run my tests, am I missing something?

veigajoao:
You should fetch the root account, which has an enormous balance, more than enough for anything you might want to do in your tests.

```
let worker: Worker<Sandbox> = workspaces::sandbox().await?;

let root = worker.root_account();

root.create_subaccount(worker, account_id)
    .initial_balance(USER_ACCOUNT_BALANCE)
    .transact()
    .await
    .unwrap()
    .unwrap()
```

shariffdev:
@frol  Let's close this issue as it has been solved.

frol:
@shariffdev This should be documented to be completely resolved.

PR #350, opened 2024-02-12
feat: create dev account with deposit on sandbox

VladasZ:
In my tests I needed to have an account with big amount of Near to test big chunks of data and pay for storage. 
I couldn't find a way to create account with more Near than hardcoded by default so I added a new method: `dev_create_account_with_deposit`.


frol:
@VladasZ See #156. That is much easier way to achieve what you have here. If you want to contribute, consider documenting the solution in #156 instead of creating extra methods.

PR #351, opened 2024-02-13
docs: a way to get account with more than 100 near on sandbox

VladasZ:
This issue: https://github.com/near/near-workspaces-rs/issues/156 and my PR: https://github.com/near/near-workspaces-rs/pull/350 indicate that it is not clear for users how to create a dev account with more than 100 Near.

Closes: https://github.com/near/near-workspaces-rs/issues/156

### DIFF
--- a/workspaces/src/network/variants.rs
+++ b/workspaces/src/network/variants.rs
@@ -78,6 +78,18 @@ where
         (id, sk)
     }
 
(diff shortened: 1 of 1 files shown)


VladasZ:
I'm not sure why test job is failing. I only added a comment and it works fine on my mac.

frol (approved):

PR #360, opened 2024-06-06
feat: Fixed `dev_deploy` to use registrar account to create top-level accounts (nearcore 1.37.0+ forbidden to create top-level accounts by non-registrar accounts)

fospring:
to fix issue: https://github.com/near/near-workspaces-rs/issues/354


fospring:
@frol pls take a look

workspaces/src/network/sandbox.rs
```diff
@@ -142,11 +146,15 @@ impl TopLevelAccountCreator for Sandbox {
     async fn create_tla_and_deploy(
         &self,
         worker: Worker<dyn Network>,
-        id: AccountId,
+        account_prefix: String,
```
frol, line 149: TLA stands for "top-level account", so this does not make sense.

The TLA account creation transaction must be signed as registrar account. Could you, please, generate a sandbox home folder and attach the genesis.json file here? It should [have `registrar_account_id`](https://github.com/near/nearcore/blob/677a59fdcc11fcd179ff4719ded4c41226e347f1/utils/mainnet-res/res/mainnet_genesis.json#L207-L210), and have it in the `"records"` together with the access key, so I hope to see if we control that account and if not, we should implement it, it we already do, just use that instead of the `root_signer`.

fospring: ok, I will try it

workspaces/Cargo.toml
```diff
@@ -39,11 +39,11 @@ near-crypto = "0.20.0"
 near-primitives = "0.20.0"
 near-jsonrpc-primitives = "0.20.0"
 near-jsonrpc-client = { version = "0.8", features = ["sandbox"] }
-near-sandbox-utils = "0.7.0"
+near-sandbox-utils = {git = "https://github.com/near/sandbox", tag = "v0.7.1"}
```
frol, line 42: Just for the context, due to the missing near-sandbox 1.39.0 pre-built binaries, I had a little detour: https://near.zulipchat.com/#narrow/stream/295306-contract-runtime/topic/neard.20sandbox.20for.20performance.20analysis

I will release near-sandbox 0.8.0 with nearcore 1.38.0 today

workspaces/src/network/sandbox.rs
fospring: I registered `registerar` account's key and initial balance in `genesis.json` and create `registerar.json` file as registerar's wallet. but two test failed, I will check

workspaces/Cargo.toml
fospring: updated

frol (changes requested):
Overall, looks good. Please, address the comments and fix CI and we are good to go

workspaces/src/network/config.rs
```diff
@@ -74,3 +75,83 @@ pub(crate) fn set_sandbox_configs(home_dir: impl AsRef<Path>) -> Result<()> {
...
+    let config = config.as_object_mut().expect("expected to be object");
+    let mut total_supply = u128::from_str(
+        config
+            .get_mut("total_supply")
+            .expect("expected exist total_supply")
+            .as_str()
+            .unwrap_or_default(),
+    )
+    .unwrap_or_default();
+    let register_amount = 10_000_000_000_000_000_000_000_000_000_u128;
```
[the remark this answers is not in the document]
fospring: update typo

workspaces/src/network/sandbox.rs
```diff
@@ -126,7 +140,7 @@ impl TopLevelAccountCreator for Sandbox {
         id: AccountId,
         sk: SecretKey,
     ) -> Result<Execution<Account>> {
-        let root_signer = self.root_signer()?;
+        let root_signer = self.register_signer()?;
```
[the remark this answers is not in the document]
fospring: typo updated

frol (approved):

PR #362, opened 2024-07-01
Testnet: error handling and add .testnet to dev_generate

AlexKushnir1:
Add error handling to url_create_account https://github.com/near/near-workspaces-rs/issues/353
Response from url_create_account method return next error: A top-level account ID AccountId("dev-20240701184705-27025377514474") can't be created by AccountId("testnet"), short top-level account IDs can only be created by AccountId("registrar")

### DIFF
--- a/workspaces/src/network/variants.rs
+++ b/workspaces/src/network/variants.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use crate::network::Info;
 use crate::result::{Execution, Result};
 use crate::rpc::client::Client;
@@ -75,6 +77,11 @@ where
     pub async fn dev_generate(&self) -> (AccountId, SecretKey) {
         let id = crate::rpc::tool::random_account_id();
         let sk = SecretKey::from_seed(KeyType::ED25519, DEV_ACCOUNT_SEED);
+        if self.info().name.eq("testnet") {
+            let id: AccountId = AccountId::from_str(format!("{}.testnet", id.as_str()).as_str())
+                .expect("could not convert string account id into AccountId");
+            return (id, sk);
+        }
         (id, sk)
     }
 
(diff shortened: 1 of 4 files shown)


workspaces/src/network/variants.rs
```diff
@@ -75,6 +77,11 @@ where
     pub async fn dev_generate(&self) -> (AccountId, SecretKey) {
         let id = crate::rpc::tool::random_account_id();
         let sk = SecretKey::from_seed(KeyType::ED25519, DEV_ACCOUNT_SEED);
+        if self.info().name.eq("testnet") {
+            let id: AccountId = AccountId::from_str(format!("{}.testnet", id.as_str()).as_str())
+                .expect("could not convert string account id into AccountId");
+            return (id, sk);
+        }
```
frol, line 84: This is not a maintainable way of addressing it. I cannot approve this PR with a workaround. Consider putting this code somewhere in the testnet dev_create_account/url_create_account implementation.

workspaces/tests/account.rs
```diff
@@ -1,10 +1,34 @@
...
+use std::str::FromStr;
+
+#[test(tokio::test)]
+async fn test_dev_account_creation() -> anyhow::Result<()> {
+    let worker: Worker<Sandbox> = near_workspaces::sandbox().await?;
+    worker.dev_create_account().await?;
+
+    let worker: Worker<Testnet> = near_workspaces::testnet().await?;
+    let (id, sk) = worker.dev_generate().await;
+    worker.create_tla(id, sk).await?.into_result()?;
```
frol, line 21: This should not pass. TLA stands for top-level account (account name without `.`, e.g. `near`, `testnet`). `create_tla` on testnet cannot succeed.

frol (changes requested):

PR #363, opened 2024-07-05
Upgrade

akorchyn:


workspaces/src/types/account.rs
```diff
@@ -393,7 +395,9 @@ pub struct AccountDetails {
     pub balance: NearToken,
     pub locked: NearToken,
     pub code_hash: CryptoHash,
-    pub storage_usage: u64,
+    pub storage_usage: StorageUsage,
+    // TODO: protocol_feature_nonrefundable_transfer_nep491 is not supported by near-cli-rs
```
frol, line 399: Is it TODO for near-workspaces-rs or for near-cli-rs?

frol: @akorchyn Protocol feature flags on nearcore creates are dropped after they get stabilized. Are you sure you need an unstable protocol feature here?

frol (approved):

PR #367, opened 2024-08-02
feat: Cacheing the nonce in case of transact_async

jaswinder6991:
- Creating a nonce cache for async transaction. 
- The nonce is stored locally and incremented only sequentially. 

Note: The PR might be a little rough around the edges in terms of best practices and thread safety. Hence, your feedback is welcome.

### DIFF
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -29,6 +29,7 @@ tokio = { version = "1", features = ["full"] }
 tokio-retry = "0.3"
 tracing = "0.1"
 url = { version = "2.2.2", features = ["serde"] }
+once_cell = "1.19.0"
 
 near-abi-client = "0.1.1"
 near-gas = { version = "0.2.5", features = ["serde", "borsh", "schemars"] }
--- a/workspaces/Cargo.toml
+++ b/workspaces/Cargo.toml
@@ -29,7 +29,6 @@ tokio = { version = "1", features = ["full"] }
 tokio-retry = "0.3"
 tracing = "0.1"
 url = { version = "2.2.2", features = ["serde"] }
-once_cell = "1.19.0"
 
 near-abi-client = "0.1.1"
 near-gas = { version = "0.2.5", features = ["serde", "borsh", "schemars"] }
--- a/workspaces/src/rpc/client.rs
+++ b/workspaces/src/rpc/client.rs
@@ -2,7 +2,6 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
-use tokio::sync::Mutex;
 
 use crate::types::NearToken;
 use near_gas::NearGas;
@@ -27,8 +26,6 @@ use near_primitives::views::{
     TxExecutionStatus,
 };
 
-use once_cell::sync::Lazy;
-
 #[cfg(feature = "experimental")]
 use {
     near_chain_configs::{GenesisConfig, ProtocolConfigView},
@@ -50,9 +47,6 @@ use crate::{Network, Worker};
 
 pub(crate) const DEFAULT_CALL_FN_GAS: NearGas = NearGas::from_tgas(10);
 pub(crate) const DEFAULT_CALL_DEPOSIT: NearToken = NearToken::from_near(0);
-
-static NONCE_CACHE: Lazy<Mutex<HashMap<(AccountId, near_crypto::PublicKey), Nonce>>> =
-    Lazy::new(|| Mutex::new(HashMap::new()));
 
 /// A client that wraps around [`JsonRpcClient`], and provides more capabilities such
 /// as retry w/ exponential backoff and utility functions for sending transactions.
@@ -507,16 +501,20 @@ async fn fetch_tx_nonce(
         let (access_key, block_hash) =
             access_key(client, account_id.clone(), public_key.clone()).await?;
 
-        // case where multiple writers end up at the same lock acquisition point and tries
-        // to overwrite the cached value that a previous writer already wrote.
-        let nonce = client
-            .access_key_nonces
-            .write()
-            .await
+        let mut nonces = client.access_key_nonces.write().await;
+        // Double-check if another thread has updated the nonce while we were fetching the access key
+        if let Some(nonce) = nonces.get(cache_key) {
+            let nonce = nonce.fetch_add(1, Ordering::SeqCst);
+            return Ok((block_hash, nonce + 1));
+        }
+
+        // Insert the new nonce into the cache
+        let nonce = nonces
             .entry(cache_key.clone())
             .or_insert_with(|| AtomicU64::new(access_key.nonce + 1))
             .fetch_max(access_key.nonce + 1, Ordering::SeqCst)
             .max(access_key.nonce + 1);
+        drop(nonces);
 
         Ok((block_hash, nonce))
     }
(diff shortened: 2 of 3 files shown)


jaswinder6991:
@race-of-sloths

race-of-sloths:
@jaswinder6991 Thank you for your contribution! Your pull request is now a part of the [Race of Sloths](https://race-of-sloths.com)!
New Sloth joined the Race! Welcome!

[<picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://badge.race-of-sloths.com/jaswinder6991?type=bot&theme=dark">
    <source media="(prefers-color-scheme: light)" srcset="https://badge.race-of-sloths.com/jaswinder6991?type=bot&theme=light">
    Shows profile picture for the author of the PR
</picture>](https://race-of-sloths.com/profile/jaswinder6991)


<details><summary>Current status: <i>waiting for scoring</i></summary>

We're waiting for maintainer to score this pull request with `@race-of-sloths score [0,1,2,3,5,8,13]` command. Alternatively, autoscoring [1,2] will be applied for this pull request

</details><details><summary>What is the Race of Sloths</summary>
<p>
Race of Sloths is a friendly competition where you can participate in challenges and compete with other open-source contributors within your normal workflow

For contributors:

- Tag @race-of-sloths inside your pull requests
- Wait for the maintainer to review and score your pull request
- Check out your position in the [Leaderboard](https://race-of-sloths.com/leaderboard)
- Keep weekly and monthly streaks to reach higher positions
- Boast your contributions with a dynamic picture of your [Profile](https://race-of-sloths.com/profile/jaswinder6991)

For maintainers:
- Score pull requests that participate in the Race of Sloths
- Engage contributors with fair scoring and fast responses so they keep their streaks
- Promote the Race to the point where the Race starts promoting you
- Grow the community of your contributors

Feel free to check [our website](https://race-of-sloths.com) for additional details!
</p>
</details> 

<details>
<summary>Bot commands</summary>

- For contributors
  - **Include a PR:** `@race-of-sloths include` to enter the Race with your PR
- For maintainers:
  - **Assign points:** `@race-of-sloths score [1/2/3/5/8/13]` to award points based on your assessment.
  - **Reject this PR:** `@race-of-sloths exclude` to send this PR back to the drawing board.
  - **Exclude repo:** `@race-of-sloths pause` to stop bot activity in this repo until `@race-of-sloths unpause` command is called

</details>

frol (changes requested):
There is already implementation for that inside `fetch_tx_nonce`: https://github.com/near/near-workspaces-rs/blob/445e86e4e73fc044962228c752fa4fc3493d24a7/workspaces/src/rpc/client.rs#L494-L503

I believe the problem is that it does not work for the first transactions since if there is no nonce cached, it will end up in the else branch too early and thus several concurrent jobs will end up querying the latest nonce and use that.

jaswinder6991:
> There is already implementation for that inside `fetch_tx_nonce`:
> 
> https://github.com/near/near-workspaces-rs/blob/445e86e4e73fc044962228c752fa4fc3493d24a7/workspaces/src/rpc/client.rs#L494-L503
> 
> I believe the problem is that it does not work for the first transactions since if there is no nonce cached, it will end up in the else branch too early and thus several concurrent jobs will end up querying the latest nonce and use that.

Nice. My B, I should have checked it. 
Anyway, do you suggest I fix the race condition in the fetch_nonce function or just take it out of here and handle it like I have done in the `send_batch_tx_async` function. 

One reason, I did not want to touch fetch_nonce was because I thought it might break other functions.

jaswinder6991:
@frol so I tried to simplify my solution by checking if another thread has already update the cache (L506):

https://github.com/near/near-workspaces-rs/blob/b28148b90bab8b056e211606b7ca12091d75b018/workspaces/src/rpc/client.rs#L497-L519

It worked fine for me locally, seems like it failed in CI tests. However it has achieved some level of concurrency. 

Do you recommend I dig deeper or this should be fine?

frol:
Thinking about this problem more, I believe that the issue is not due to the nonce, the implementation in master doesn’t have a race

https://github.com/near/near-workspaces-rs/blob/445e86e4e73fc044962228c752fa4fc3493d24a7/workspaces/src/rpc/client.rs#L510-L519

This is the right way to handle it.

There are two other problems:

1. Making 50 parallel calls means that nonces could be assigned to these requests randomly (it is NOT msg0 = nonce+1, msg1 = nonce+2, … msg49 = nonce+50. Thus, given that blockchain will order the transactions by nonce, you may easily end up with msg17 having the highest nonce and executed last
2. Given that the transactions do not arrive all at once, you may end up with a transaction with nonce+50 to be executed at block height 10, and all the transactions that have nonces lower that nonce+50 and arrived late for that block will be discarded (not applied at block height 11)

jaswinder6991:
Hi @frol thanks for your helpful insight. I understand it now too. 

Seems like the nonce never had the issue. I went on the wrong track and ofcourse the test is incorrect. 
Also, the telegram query in question was about sequential async txs, which work fine, here is a test for that: 

```rust
use anyhow::Result;
use serde_json::json;
use std::time::Duration;

const STATUS_MSG_CONTRACT: &[u8] = include_bytes!("../../examples/res/status_message.wasm");

#[tokio::test]
async fn test_nonce_caching_sequential_async_no_wait() -> Result<()> {
    let worker = near_workspaces::sandbox().await?;
    let contract = worker.dev_deploy(STATUS_MSG_CONTRACT).await?;
    let account = worker.dev_create_account().await?;

    // Get the initial nonce
    let initial_nonce = worker
        .view_access_key(account.id(), &account.secret_key().public_key())
        .await?
        .nonce;

    const NUM_TRANSACTIONS: usize = 10;
    let mut transaction_statuses = Vec::with_capacity(NUM_TRANSACTIONS);

    // Send transactions sequentially without waiting for completion
    for i in 0..NUM_TRANSACTIONS {
        let msg = format!("msg{}", i);
        let result = account
            .call(&contract.id(), "set_status")
            .args_json(json!({ "message": msg }))
            .transact_async()
            .await?;

        transaction_statuses.push(result);

        // Small delay to simulate some processing time between transactions
        //tokio::time::sleep(Duration::from_millis(10)).await;
    }

    // Now wait for all transactions to complete
    for (i, status) in transaction_statuses.into_iter().enumerate() {
        loop {
            match status.status().await? {
                std::task::Poll::Ready(outcome) => {
                    if outcome.is_success() {
                        println!("Transaction {} completed successfully", i);
                        break;
                    } else {
                        return Err(anyhow::anyhow!("Transaction {} failed: {:?}", i, outcome));
                    }
                }
                std::task::Poll::Pending => {
                    tokio::time::sleep(Duration::from_millis(100)).await;
                }
            }
        }
    }

    // Get the final nonce
    let final_nonce = worker
        .view_access_key(account.id(), &account.secret_key().public_key())
        .await?
        .nonce;

    // Verify that the nonce increased by exactly the number of transactions
    assert_eq!(
        final_nonce - initial_nonce,
        NUM_TRANSACTIONS as u64,
        "Nonce did not increment correctly"
    );

    Ok(())
}
```

Going back to the original query on Telegram:

> Hey, does near-workspaces have a way to send txs sequentially (with proper nonce) in an async execution context?
If I do this:
> ``` let users = try_join_all((0..user_amount).map(|_| worker.dev_create_account())).await?;```
> then I randomly get invalid nonce error

I think it is more of an account creation issue in async context. 

I will close this PR now. 
As an aside, I apologise for pulling you into this, I appreciate your valuable time.

PR #408, opened 2025-03-14
chore: remove repetitive words

jasmyhigh:


akorchyn (approved):
Thanks for the contribution ❤️

@race-of-sloth score 2

PR #417, opened 2025-05-15
ci: pin to 1.86.0 max

dj8yfo:


dj8yfo:
@race-of-sloths

akorchyn (approved):

akorchyn:
@race-of-sloths score 5

PR #423, opened 2025-08-15
chore: update near-* dependencies to 0.31

vsavchyn-dev:


frol (approved):
@race-of-sloths score 2

